PR #26, opened 2019-05-31
begin work on integrating the Kubernetes Rest API for various functions

bfitzpat:

### Things that are working

* With the proxy enabled, we can now: populate the Integrations view, remove (undeploy) a running integration, and start (deploy) a new integration for Java, XML, Javascript, and Groovy camel files
* Auto-refresh now works consistently across kubectl and rest and provides more meaningful results in the Camel-K output if issues arise (such as seeing zero integrations but having a successful refresh or if the proxy is not available)
* Removed automatically streaming output from running integrations to the Camel-K output channel in favor of having a new menu (View log snapshot for Camel-K Integration) in the Camel-K Integrations view. This works for both the proxy-based and executable-based approaches we have currently
* Added a new command (F1 or Ctrl+Shift+P to open command palette) to start the local kubernetes proxy server (kubectl proxy –-port=8000). The proxy will be shut down when the user closes the VS Code instance
* Output channels created for streaming output from running integrations now update once opened until the integration is removed.


src/CamelKJSONUtils.ts
```diff
@@ -0,0 +1,147 @@
...
+	let server = vscode.workspace.getConfiguration().get(proxyURLSetting);
+	let namespace = vscode.workspace.getConfiguration().get(proxyNamespaceSetting);
+	let outputUrl = server + "/api/v1/namespaces/" + namespace + "/pods";
+	return outputUrl;
+}
+
+function cleanText(text:string) : Promise<string> {
+	return new Promise( (resolve, reject) => {
+		if (text) {
+			let reg = new RegExp('(?:\\\")*');
```
apupier: why is it needed to c"clean" text with specific code? It is a non-standard encoded text? where is the encoding done on the other side so that we can check that we covered all cases?

README.md
```diff
@@ -65,9 +68,28 @@ The view has a "Refresh" button that can be used to manually trigger a refresh o
...
+* Use Proxy - This setting determines whether the Camel-K Integrations view retrieves the list of running integrations via the local 'kubectl' application or via the Kubernetes Rest API and calls through the Proxy URL/Namespace combination above. The ultimate URL becomes [proxyurl]/apis/camel.apache.org/v1alpha1/namespaces/[namespace]/integrations. For now this is only used when retrieving the integrations list, but it will also be used when starting and stopping integrations when the "Use Proxy" setting is set to true.
+
+## Starting a local Kubernetes proxy
+
+We have created a new command available in the command palette (Ctrl+Shift+P or F1) called "Camel-K: Start the kubectl proxy server." This will start a new Kubernetes proxy using the Minikube executable ('kubectl proxy --port=8000) and refreshes the Camel-K Integrations view immediately. This command creates a local proxy at http://localhost:8000, which is the default value for Proxy URL in the extension settings.
```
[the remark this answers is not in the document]
bfitzpat: No, there is no auto-install of Minikube that goes with this. It would be too costly resource-wise and I think not a good decision. This is more for a local development case and I'll add that as a note in the readme.

package.json
```diff
@@ -30,7 +30,7 @@
 		"Tool"
 	],
 	"engines": {
-		"vscode": "^1.33.0"
+		"vscode": "^1.29.0"
```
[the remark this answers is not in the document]
bfitzpat: I was just dropping it to be consistent with the other vscode extensions we have in development. I can bump it back to 1.33. Thanks

src/CamelKJSONUtils.ts
```diff
@@ -0,0 +1,147 @@
...
+			reject();
+		}
+	});
+}
+
+export function createCamelKDeployJSON( name:string, fileContent:string, fileName:string) : Promise<string> {
+	return new Promise( (resolve) => {
+		let content = {
+			"kind":"Integration",
+			"apiVersion":"camel.apache.org/v1alpha1",
```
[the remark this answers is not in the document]
bfitzpat: Great question. I have asked the powers that be about the future of this version.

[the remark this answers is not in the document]
bfitzpat: In the meantime, I'll spin off the "v1alpha1" as a constant so I can change it one place when it does change.

src/extension.ts
```diff
@@ -85,60 +338,64 @@ function startIntegration(context: vscode.ExtensionContext): Promise<string> {
...
-		// Otherwise we can't shut it down properly from this direction and have
-		// to use the "remove" method from the Integrations view.
-
-		let commandString = 'kamel delete "' + filename + '"';
-		console.log('Command string: ' + commandString);
-		child_process.exec(commandString, { cwd : absoluteRoot}, (error, stdout, stderr) => {
-			if (error) {
-				console.error(`exec error: ${error}`);
+// use kubernetes rest to post new integration
+function callKamel2(): Promise<boolean> {
```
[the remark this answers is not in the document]
bfitzpat: Was just a holdover from copy/paste change I did a month ago. I will rename callKamel and callKamel2 to be more explicit in their functionality.

src/extension.ts
```diff
@@ -187,4 +427,55 @@ function callKamel(context: vscode.ExtensionContext): Promise<boolean> {
...
+			return podName;
+		}
+	}
+	return '';
+}
+
+// start a new local Kubernetes proxy using kubectl
+function startKubeProxy(): Promise<string> {
+	return new Promise( (resolve, reject) => {
+		let commandString = 'kubectl proxy --port=8000';
```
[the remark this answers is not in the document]
bfitzpat: Good thought. Created https://github.com/camel-tooling/vscode-camelk/issues/34

src/CamelKJSONUtils.ts
bfitzpat: When wrapping the file contents in the JSON for deployment to Kubernetes, I discovered some interesting issues when using JSON.stringify that had to be corrected, thus the RegEx dancing. See lines 75 and 102 where the cleanText function is called. This is just a sanitizing pass to ensure that we aren't doubling up the slashes, which was happening.

src/CamelKJSONUtils.ts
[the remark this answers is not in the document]
bfitzpat: The API will change eventually to v1 instead of v1alpha1, but there will be a transition period per Nicola. The API is here - https://github.com/apache/camel-k/tree/master/pkg/apis/camel

src/CamelKJSONUtils.ts
apupier: I still don't get why JSON.stringify is returning invalid strings. it is a bug on what the server is returning? or on the way we are decoding the response?

package.json
```diff
@@ -107,19 +140,24 @@
...
+		"vscode": "^1.1.34"
 	},
 	"dependencies": {
+		"@types/nock": "^10.0.3",
+		"@types/request-promise": "^4.1.44",
 		"chai": "^4.2.0",
 		"child_process": "^1.0.2",
 		"path": "^0.12.7",
+		"request": "^2.88.0",
+		"request-promise": "^4.2.4",
```
apupier: based on Paolo comment on using fetch instead of request-promise would worth looking at it but I think it is fine if we are looking at it in a second iteration (also fine if you modify it directly in this PR, it will just delay it a bit)

README.md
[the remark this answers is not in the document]
apupier: I'm not sure that I was clear enough.

I did not meant to have a Kubernetes instance started locally using Minikube. I meant, that we need to have the kubectl binary available on the command-line to have the proxy creation which is called through `kubectl proxy --port=8000`
I looked to the Kubernetes VS Code extension and they are [requiring to have kubectl installed locally](https://code.visualstudio.com/docs/azure/kubernetes#_before-you-begin), so even if it is not the best user experience, i think we can expect that users playing with Camel-K will also have it.

src/CamelKJSONUtils.ts
bfitzpat: For some reason I saw all sorts of doubling of slashes between grabbing the file content and using JSON. Stringify -- WHY that happened, I don't know. But this little cleanup pass does the trick and I am able to package Groovy, Java, JavaScript, and XML files in the YAML format they need for deploying an integration.

package.json
bfitzpat: I got things working with curl, but will definitely look into the fetch approach to replace request-promise after we get things mostly nailed down. I'd rather do that in a separate PR.

src/CamelKJSONUtils.ts
apupier: @lburgazzoli Do you have an idea why the provided json contains several double-slash, which seems to be invalid as it needs to be processed on client side. is it a specific encoding which needs to be decoded? or is ti a manual task which is doubling the escape so that we can dedouble them also in our codebase?

lburgazzoli: sorry, which json ?

apupier: hum i thought that it would have been somethign specific to the Camel-K call.
by re-reading code, it seems that ti is not. In fact, it seems that it is because Json.stringify is called instead of JSON.parse or simply reusing the reference (if I understand well this Stackoverflow answer https://stackoverflow.com/questions/16507858/json-stringify-escapes-double-quotes-every-time-when-stringified )

bfitzpat: Discovered we no longer needed the cleanText method at all. After removing the JSON.stringify from the file text, all the extra slashes went away by themselves. The cleanText just ended up being unnecessary in the end.

package.json
```diff
@@ -59,19 +85,22 @@
 			},
 			{
 				"command": "camelk.integrations.remove",
-				"title": "Remove Integration"
+				"title": "Remove Camel-K Integration"
+			},
+			{
+				"command": "camelk.integrations.log",
+				"title": "View log for Camel-K Integration"
```
apupier: in Kubernetes extension, the naming used is:
"Show logs" if provide value at a single point of time
"Follow logs" to have logs updated.

I think it would be nice to reuse the same naming

package.json
```diff
@@ -59,19 +85,22 @@
...
+			},
+			{
+				"command": "camel.integrations.startproxy",
```
apupier: I think there is a typo, would be best with "camelk.integrations.startproxy" instead of "camel.integrations.startproxy"

package.json
```diff
@@ -59,19 +85,22 @@
...
+				"title": "Camel-K: Start the kubectl proxy server"
```
apupier: provided a follow-up enhancement https://github.com/camel-tooling/vscode-camelk/issues/38 to guide user

src/extension.ts
```diff
@@ -20,62 +20,322 @@ import * as vscode from 'vscode';
...
+			} else {
+				utils.shareMessage(mainOutputChannel, 'Removing ' + integrationName + ' via Kamel executable Delete');
+				let commandString = 'kamel delete "' + integrationName + '"';
+				child_process.exec(commandString, (error, stdout, stderr) => {
+					if (error) {
+						utils.shareMessage(mainOutputChannel, `exec error: ${error}`);
+						return;
+					}
+					if (stdout) {
+						// utils.shareMessage(mainOutputChannel, `stdout: ${stdout}`);
```
apupier: why it is commented out?
Either remove the if and the commented code or uncomment.

src/extension.ts
```diff
@@ -20,62 +20,322 @@ import * as vscode from 'vscode';
...
+			await utils.pingTheURL(podsURL).catch( (error) =>  {
+				reject(error);
+			});
+			let podOutputChannel = getOutputChannel(podName);
+			podOutputChannel.show();
+			let commandString = `curl -v -H "Accept: application/json, */*" ${podsURL}?follow=true --insecure -N`;
+			await utils.delay(1000);
+			let runKamel = child_process.spawn(curlCommand, [curlOption, commandString]);
+			runKamel.stdout.on('data', function (data) {
+				podOutputChannel.append(`${data}`);
```
apupier: maybe in a next iteration but it seems that Kubernetes extension is providing a nice Log viewer as a view or an editor. it might be better than using the podOutputChannel (and still cannot test but it seems that in case of several pod, the log will be mixed in it, isn't it?)

package.json
bfitzpat: Will change to "Follow log for Camel-K Integration"

src/extension.ts
bfitzpat: So far as I know, we can't rely on elements from other extensions -- has that changed?

apupier: you're right, I tend to forget that VS Code is so powerless...
Then maybe in a next iteration, duplicate the log editor code (or find a way to extract it as a normal javascript extension and submit it)

bfitzpat: Created https://github.com/camel-tooling/vscode-camelk/issues/39

Issue #35, opened 2019-07-04
Move test dependencies to devDependencies


apupier:
I think there is:
chai, sinon and sinon-chai

bfitzpat:
PR was merged

apupier:
we should not close the issue inside camel-tooling ourself, but move them in "QE" column of the work planning board https://github.com/orgs/camel-tooling/projects/1

djelinek:
verified with master branch

PR #37, opened 2019-07-04
Move test dependencies to devDependencies #35

apupier:
it is saving space in the final vsix binary


bfitzpat:
Need to rebase this one on master, but I think it's a good fix.

PR #43, opened 2019-07-10
fixing submitter list so we can publish vscode-camelk via jenkins

bfitzpat:

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -1,8 +1,23 @@
-# How to release
+# How to provide a new version on VS Code Marketplace
 
-- Create a tag and push it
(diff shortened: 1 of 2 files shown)


bfitzpat:
> 
> 
> will need to update also the instruction to publish a release in the contributing.md too but can be done in another PR

I can add it here - is it in the atlasmap contributing.md file?

apupier:
> > will need to update also the instruction to publish a release in the contributing.md too but can be done in another PR
> 
> I can add it here - is it in the atlasmap contributing.md file?

yes https://github.com/jboss-fuse/vscode-atlasmap/blob/master/Contributing.md

Issue #45, opened 2019-07-17
Find better way to re-publish routes


bfitzpat:
Currently we have the ability to:
* Deploy a new integration
* View the log for the running integration
* Delete the integration (which closes the Output for the log if one is open)

But what happens in a slightly more complex case? Let's say we want to:
* Deploy an integration
* View the log for the running integration
* Update the route for the integration and re-deploy (new)
* Delete the integration

Right now if we update the integration and re-deploy we gain additional artifacts. Out of date Output channels for logs for pods that are shut down remain open. 

We need to decide what the best approach is here because this will be a common use case.

We are not listening for changes to the file currently because deploying via the executable operates in a slightly different way than deploying via the Rest APIs in Kubernetes. Since we are managing the deployment JSON, we are not taking advantage of the connection I have seen between the running integration just magically *updating* when saved in the file system. And we don't really have an explicit link between the file name and the deployed integration, though that information is available in the deployment descriptor for the integration.

Is it enough to simply be smarter about re-deploying the integration and cleaning up the older Output channels?

apupier:
it is the kind of issues that odo and OpenShift Connector VS Code extension project are working on. Would be interesting to reach them, and/or to reuse what they are doing.

bfitzpat:
Approach I'm considering now...

* take the filename of the camel file we're running via Camel-K
* remove the extension
* change to Kebab case (similar to https://github.com/apache/camel-k/blob/master/pkg/util/kubernetes/sanitize.go, which is what they're doing)
* look for the pod that uses that integration name (already have code for this)
* remove the old output channel
* proceed with deployment -- it seems that camel-k is smart enough to remove the old version first

bfitzpat:
Fun fact, I've discovered an issue with the latest version of Kubernetes and Minikube with Camel-K...
```
I used to be able to get the pod name for a running integration via: kubectl get pods | grep {integrationName}
but now with minikube 1.2.0 and kubernetes version 1.15.0 I am not seeing any of the deployed integrations as pods any longer
I see them in the custom resources/integrations list
but I don't seem to be able to get the log for a custom resource
this was working with minikube 1.1.1 and kubernetes version 1.12.0 and seems to be new behavior
```

bfitzpat:
Found a good example of the Kebab case fix for Javascript - https://gist.github.com/thevangelist/8ff91bac947018c9f3bfaad6487fa149

bfitzpat:
[...]

PR #46, opened 2019-07-19
issue #45 - removing output log if it exists -also did some cleanup t…

bfitzpat:
…o avoid repeated code

### DIFF
--- a/src/CamelKJSONUtils.ts
+++ b/src/CamelKJSONUtils.ts
@@ -17,6 +17,7 @@
 import * as fs from 'fs';
 import * as vscode from 'vscode';
 import * as http from "http";
+import * as child_process from 'child_process';
 
 export const proxyURLSetting = 'camelk.integrations.proxyURL';
 export const proxyNamespaceSetting = 'camelk.integrations.proxyNamespace';
@@ -119,10 +120,36 @@ export async function pingKubernetes() : Promise<string> {
 	});
 }
 
+export async function pingKamel() : Promise<any> {
+	return new Promise( async (resolve, reject) => {
+		const pingKamelCommand = "kamel get";
+		let runKubectl = child_process.exec(pingKamelCommand);
+		runKubectl.stdout.on('data', function (data) {
+			let output : string = data as string;
+			resolve(output);
+			return;
+		});
+		runKubectl.stderr.on('data', function (data) {
+			let error : string = data as string;
+			reject(new Error(error));
+			return;
+		});
+	}); 
+}
+
 export function shareMessage(outputChannel: vscode.OutputChannel, msg:string) {
 	if (outputChannel) {
-		outputChannel.append(msg + '\n\n');
+		outputChannel.append(msg + '\n');
 	} else {
 		console.log('[' + msg + ']');
 	}
+}
+
+export function toKebabCase (str : string) {
+	return str.replace(/([A-Z])([A-Z])/g, '$1-$2')
+		.replace(/([a-z])([A-Z])/g, '$1-$2')
+		.replace(/[\s_]+/g, '-')
+		.replace(/^[-]+/g, '')
+		.replace(/[-]$/, '')
+		.toLowerCase() ;
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to the "vscode-camelk" extension will be documented in this 
 
 ## 0.0.5
 
+- Added status bar indicator when Camel-K integrations view is refreshing
+- Added test for kebab case based on Camel-K go test
+- Added check to see if the kamel executable is available when not using proxy
+- Fixed issue with right-click menu not correctly finding selected file
+
 ## 0.0.4
 
 - Added Camel-K Integration Settings where users can specify the proxy url and namespace to use the Kubernetes Rest APIs instead of the local 'kubectl' utility.
--- a/package-lock.json
+++ b/package-lock.json
@@ -652,9 +652,9 @@
             "dev": true
         },
         "lodash": {
-            "version": "4.17.11",
-            "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.11.tgz",
-            "integrity": "sha512-cQKh8igo5QUhZ7lg38DYWAxMvjSAKG0A8wGSVimP07SIUEK2UO+arSRKbRZWtelMtN5V0Hkwh5ryOto/SshYIg=="
+            "version": "4.17.15",
+            "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
+            "integrity": "sha512-8xOcRHvCjnocdS5cpwXQXVzmmh5e5+saE2QGoeQmbKmRS6J3VQppPOIt0MnmE+4xlZoumy0GPG0D0MVIQbNA1A=="
         },
         "lolex": {
             "version": "4.0.1",
--- a/package.json
+++ b/package.json
@@ -134,8 +134,9 @@
 		}
 	},
 	"scripts": {
-		"vscode:prepublish": "npm run compile",
+		"vscode:prepublish": "npm run clean && npm run compile",
 		"compile": "tsc -p ./",
+		"clean": "rm -rf out || rmdir out /s /q",
 		"watch": "tsc -watch -p ./",
 		"postinstall": "node ./node_modules/vscode/bin/install",
 		"test": "npm run compile && node ./node_modules/vscode/bin/test"
--- a/src/CamelKNodeProvider.ts
+++ b/src/CamelKNodeProvider.ts
@@ -83,6 +83,8 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 	// trigger a refresh event in VSCode
 	public async refresh(): Promise<void> {
 		let oldCount = this.treeNodes.length;
+		extension.myStatusBarItem.text = `Refreshing Camel-K Integrations view`;
+		extension.myStatusBarItem.show();
 		this.resetList();
 		let inaccessible = false;
 		if (this.retrieveIntegrations) {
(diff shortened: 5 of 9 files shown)


bfitzpat:
Currently the PR removes the output channel if the integration has already been published. (It also cleaned up some code, removing some duplicated areas and streamlining the approach.)

This is a work in progress and I want to have it also re-open the new output channel if the integration is being re-published.

Might also add a new setting that automatically opens the output channel for a new integration.

src/CamelKJSONUtils.ts
```diff
@@ -126,3 +126,30 @@ export function shareMessage(outputChannel: vscode.OutputChannel, msg:string) {
...
+
+export function toKebabCase (str : string) {
+	return str.replace(/([A-Z])([A-Z])/g, '$1-$2')
+		.replace(/([a-z])([A-Z])/g, '$1-$2')
+		.replace(/[\s_]+/g, '-')            
+		.toLowerCase() ;
+}
+
+// if a camel route file is selected in the explorer, it is opened in an editor.
+// if the user right-clicks to start a new integration, the same thing occurs.
```
apupier: if I right-click, the file is not selected for me so it will use the one that is already opened.
For instance, If I close all editors and view except the explorer. I right-click on a camel-k Groovy file and choose "Start Camel-K Integration", there is an undefined activeTextEditor:
image

apupier: there should be a callback to provide the file when the command is called, see https://stackoverflow.com/questions/51961457/how-to-get-file-name-or-path-in-vscode-extension-when-user-right-click-on-file-i (but cannot find it clear in official documentation https://code.visualstudio.com/api/references/vscode-api#commands )

src/CamelKJSONUtils.ts
```diff
@@ -126,3 +126,30 @@ export function shareMessage(outputChannel: vscode.OutputChannel, msg:string) {
 		console.log('[' + msg + ']');
 	}
 }
+
+export function toKebabCase (str : string) {
```
apupier: I really see this function to be a weak point for the future. if the naming pattern is modified, we need something to detect it.
I can imagine several possibilities:
- provide integration test with a running Camel-K instance and tests that are "passing through" this method
- have a test upstream which is forcing them to notify us if something is changing (might not be easy to negociate)
- ask for library able to do it
- have a test pulling the camel-K code to test that it has not changed.

src/CamelKJSONUtils.ts
bfitzpat: I'll investigate that. It was NOT very well documented how to get that

bfitzpat: Fixed it in this PR

src/CamelKJSONUtils.ts
bfitzpat: Added a test locally that mimics what the Camel-K test is doing upstream at https://github.com/apache/camel-k/blob/master/pkg/util/kubernetes/sanitize_test.go

I think this is the best we can do in the short term. We can investigate further options down the line.

bfitzpat:
Here's a question however... when I run the tests locally I see "17 passing (2s)" and when we run upstream I see "11 passing (2s)" - are we missing some tests in the travis configuration I wonder?

src/CamelKJSONUtils.ts
apupier: I think the function getSelectedEditorFile can be completely removed now

src/CamelKJSONUtils.ts
apupier: the added test is checking that vscode-camel-k will be compatible with the current implementation of camel-k. It won't detect if Camel-K is changing the implementation. We still have the weak point.
I really think that we should add a way to detect that the camel-k implementation has not changed.

src/CamelKJSONUtils.ts
apupier: (and it is a lot cleaner with the callback :-) )

apupier:
> Here's a question however... when I run the tests locally I see "17 passing (2s)" and when we run upstream I see "12 passing (2s)" - are we missing some tests in the travis configuration I wonder?

I have 12 locally.
There are good chances that some tests remained in out/test folders while you developed them (moved/renamed). Please try to clean the folder and relaunch.

src/CamelKJSONUtils.ts
bfitzpat: I have added a new test that is a brute force method of ensuring that sanitize.go hasn't changed upstream on us.

bfitzpat:
Test issue has been fixed. Also added a "clean" option to package.json to ensure I don't run into this going forward.

bfitzpat:
> I'm wondering what is happening when:
>     * two files with same name are in different folders?
>     * a file has been renamed between the two deployment attempts?
>     * two files with different names but resulting in same "kebabcase" naming?

Those are interesting cases we should definitely ask about. I'm not sure we can do more than test locally and report the changes since we don't have a way to write integration tests with those cases at this time. We could report upstream to see if they are testing against them however.

> If there is no data loss, I think we can consider to implement these usecases in another iteration and report issues only.

We can at least test them in a local working environment and report issues as they arise.

> isn't the status bar can be handled in a different PR? it will allow to have smaller PR to ease review and merge.

It certainly *could* be added in another PR, but it's a simple enough change that it seems overkill to pull out the code, wait for this PR to be merged, and fix it in a separate PR for a total of less than approximately 10 lines of code.

bfitzpat:
It seems that stashing the file in a temporary folder may not work for Travis ci

```
 14 passing (1s)
  1 failing
  1) ensure that the upstream kubernetes.go sanitize in camel-k have not changed since we checked last
       test to see if the sanitize.go file has changed since we stashed it:
     Error: ENOENT: no such file or directory, open '/tmp/sanitize.go'
      at Object.openSync (fs.js:436:3)
      at Object.module.(anonymous function) [as openSync] (ELECTRON_ASAR.js:160:31)
      at Object.readFileSync (fs.js:341:35)
      at Object.exports.wrapFsWithAsar.fs.readFileSync (ELECTRON_ASAR.js:580:40)
      at Context.<anonymous> (src/test/gocamelkkubernetes.test.ts:55:17)
      at Generator.next (<anonymous>)
      at __awaiter (out/test/gocamelkkubernetes.test.js:23:71)
      at new Promise (<anonymous>)
      at __awaiter (out/test/gocamelkkubernetes.test.js:19:12)
      at Context.<anonymous> (out/test/gocamelkkubernetes.test.js:56:20)
```

src/CamelKJSONUtils.ts
apupier: seems a good way to detect potential desynchronization. (agree a bit bruteforce)

apupier:
(by the way the branch has been created again on the camel-tooling repo, which is why there are 2 Travis job triggered. It will be really great to find how to reconfigure your local repo to push to your own repo and not the camel-tooling one)

apupier:
> It seems that stashing the file in a temporary folder may not work for Travis ci

it is due to the test execution order:

```
 ensure that the upstream kubernetes.go sanitize in camel-k have not changed since we checked last
curl -o -L sanitize.go https://raw.githubusercontent.com/apache/camel-k/master/pkg/util/kubernetes/sanitize.go
      ✓ test that we can get the sanitize.go file from the camel-k git repository: 11ms
      1) test to see if the sanitize.go file has changed since we stashed it
      ✓ see if we can call the sanitize.go functions -- skipped: 0ms
```

apupier:
fixed the failing test. See my 5 commits to check required modifcations.

now awaiting results of manual testing of the use cases https://github.com/camel-tooling/vscode-camelk/pull/46#issuecomment-514395475 (which i still cannot do :'( )

bfitzpat:
@apupier thanks for fixing the failing test issue. will do the manual testing today, though I don't believe it should hold up this PR. All of those cases should be tested with the runtime executable itself (manual deployment, not thru the tooling).

* two files with same name are in different folders? (i.e. kamel run /my/path1/testroute.groovy and kamel run /my/path2/testroute.groovy)
* a file has been renamed between the two deployment attempts? (in this case I suspect it would simply deploy as a completely separate integration name i.e. kamel run /path/testroute.groovy and kamel run /path/TestRoute.groovy)
* two files with different names but resulting in same "kebabcase" naming? (i.e. kamel run /path/TestRoute.groovy and /path/test-route.groovy I suspect would collide)

All of those cases would indicate issues upstream from us, not with the tooling itself. We may be able to provide some sanity checking for such cases, but that would definitely be a separate PR.

WDYT @apupier ?

apupier:
it depends how it is failing.
i can imagien so many different scenarios, for instance with two files with same name in differen tfolders:
- first deployment
- follow log
- second deployment of the second file
-- runtime is able to handle it and create another name for it (no idea on the behavior currently so this is a guess but that's why we need to check)
- the previous log will be deleted (I guess based on what I understand from the code) although it shouldn't

bfitzpat:
Two routes with the same name in different paths behaves as we suspected:
```
$ kamel run ./folder1/myroute.groovy
integration "myroute" created
$ kamel run ./folder2/myroute.groovy
integration "myroute" updated
```

We see the same behavior using the vscode-camelk extension currently.

bfitzpat:
Renaming a route also behaves as we suspected.
```
$ kamel run MyRouteToRename.groovy
integration "my-route-to-rename" created
$ mv MyRouteToRename.groovy MyRouteRenamed.groovy
$ kamel run MyRouteRenamed.groovy
integration "my-route-renamed" created
```

It does not update the running integration if you rename it.

bfitzpat:
And lastly... two names that end up with same "kebabcase"...
```
$ ls
my-route.groovy  MyRoute.groovy
$ kamel run MyRoute.groovy
integration "my-route" created
$ kamel run my-route.groovy
integration "my-route" updated
```

A rose by any other name... is still a rose. :)

bfitzpat:
I created https://github.com/apache/camel-k/issues/870 upstream to relay these cases back to the Camel-K team.

apupier:
ok seems we have clear of outputchannel coherent with the runtime behavior.

Issue #25, opened 2019-05-30
Add a preference to turn off the "--dev" flag when starting new integrations


bfitzpat:
Currently we append the "--dev" flag to each integration that gets started. It would be nice if that was configurable and could be turned off in the settings for the extension.

The code currently uses the flag here - https://github.com/camel-tooling/vscode-camelk/blob/a992673ca1e73409603930ca9d5f478d9e767178/src/extension.ts#L159

bfitzpat:
Created as part of https://github.com/camel-tooling/vscode-camelk/pull/47

bfitzpat:
Revised our approach to --dev for this release. For now we will remove --dev support and handle it better here - https://github.com/camel-tooling/vscode-camelk/issues/56

bfitzpat:
Closing this issue in favor of https://github.com/camel-tooling/vscode-camelk/issues/56

PR #47, opened 2019-07-25
issue #25 - removing --dev setting for camel-k executable deployment and improving proxy url support

bfitzpat:

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,10 +45,22 @@
 			{
 				"title": "Camel-K Integration Settings",
 				"properties": {
+					"camelk.integrations.useProxy": {
+						"type": "boolean",
+						"default": false,
+						"description": "Always use proxy: If true, always use the proxy URL instead of Minikube",
+						"scope": "window"
+					},
 					"camelk.integrations.proxyURL": {
 						"type": "string",
-						"default": "http://localhost:8000",
-						"description": "Proxy configuration: URL to the running Kube server with accessible port",
+						"default": "http://localhost",
+						"description": "Proxy configuration: URL to the running Kube server (to be combined with port)",
+						"scope": "window"
+					},
+					"camelk.integrations.proxyPort": {
+						"type": "number",
+						"default": 8000,
+						"description": "Proxy configuration: Port to the running Kube server (combined with proxy URL)",
 						"scope": "window"
 					},
 					"camelk.integrations.proxyNamespace": {
@@ -59,12 +71,6 @@
 						],
 						"default": "default",
 						"description": "Proxy namespace: Namespace to use for accessing the Camel-K integrations",
-						"scope": "window"
-					},
-					"camelk.integrations.useProxy": {
-						"type": "boolean",
-						"default": false,
-						"description": "Always use proxy: If true, always use the proxy URL instead of Minikube",
 						"scope": "window"
 					}
 				}
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -27,10 +27,11 @@ import {platform} from 'os';
 export let mainOutputChannel: vscode.OutputChannel;
 export let myStatusBarItem: vscode.StatusBarItem;
 let camelKIntegrationsProvider = new CamelKNodeProvider();
-let useProxy : boolean = false;
+let useProxy : boolean;
 let outputChannelMap : Map<string, vscode.OutputChannel>;
 let curlCommand : string = '/bin/sh';
 let curlOption : string = '-c';
+let proxyPort : number;
 
 // This extension offers basic integration with Camel-K (https://github.com/apache/camel-k) on two fronts.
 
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -61,10 +73,10 @@
 						"description": "Proxy namespace: Namespace to use for accessing the Camel-K integrations",
 						"scope": "window"
 					},
-					"camelk.integrations.useProxy": {
+					"camelk.integrations.useDevFlag": {
 						"type": "boolean",
-						"default": false,
-						"description": "Always use proxy: If true, always use the proxy URL instead of Minikube",
+						"default": true,
+						"description": "Always use --dev flag: If true, always use the --dev flag with the Camel-K executable",
```
apupier: what do you think to add a bit more details on what the --dev flag is doing? or at least pointing to Camel-K documentation on te subject.

src/extension.ts
```diff
@@ -422,7 +439,12 @@ function createNewIntegration(context: vscode.Uri): Promise<boolean> {
 				console.error(error); 
 			});
 
-		let commandString = 'kamel run --dev "' + absoluteRoot + '"';
+		let commandStringRoot = 'kamel run';
+		if (useDevFlag) {
+			commandStringRoot = commandStringRoot + ' --dev ';
```
apupier: what about writing it commandStringRoot += ' --dev '; ?
I prefer this one as it is shorter. I find it more easy to read also, i don' thave to check that the first argument is written exactly the same than the assignment.

src/extension.ts
```diff
@@ -422,7 +439,12 @@ function createNewIntegration(context: vscode.Uri): Promise<boolean> {
...
+		}
+		let commandString = commandStringRoot + '"' + absoluteRoot + '"';
```
apupier: what about writing it commandStringRoot += '"' + absoluteRoot + '"'; ?

package.json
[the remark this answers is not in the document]
bfitzpat: Correct. So far as I can tell there's no way to indicate --dev mode using the proxy.

src/extension.ts
```diff
@@ -31,6 +31,8 @@ let useProxy : boolean = false;
 let outputChannelMap : Map<string, vscode.OutputChannel>;
 let curlCommand : string = '/bin/sh';
 let curlOption : string = '-c';
+let useDevFlag : boolean = true;
+let proxyPort : number = 8000;
```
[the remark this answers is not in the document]
bfitzpat: That's a good catch!

package.json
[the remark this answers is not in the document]
apupier: ok, maybe worth being more explicit then in the description then. Something like:
`--dev is available only if "useProxy" is set to "false"`

bfitzpat: I'm wondering if we need the --dev flag since we're handling things a little differently. https://github.com/apache/camel-k#running-integrations-in-dev-mode-for-fast-feedback

We're deploying and then getting the log later. --dev seems a bit out of that mode and not what we're doing. What do you think @apupier ?

apupier: if the --dev is effectively automatically redeploying modifications on save, i think it is an interesting behavior. Maybe we should use in UI something like "automatic redeploy on save" instead of mentioning the --dev which is used under the cover as it is only the log which is not shown automatically, right?
And just mention that automatic redeploy is working only when not using "proxyMode"

bfitzpat: Back when I started, we had everything in one output channel. When that happened, we streamed output from the --dev deployed integration and it worked as is described. As soon as we spun out the logs into separate output channels, that was no longer the case. So we're not really using the --dev option as designed and it's pretty much being ignored in the tooling.

apupier: you mean that the automatic redeployment is not working? (I think it is the interesting part of the --dev option)
If it i snot working, we can completely remove the --dev
If it is working, i think it is a cool feature and that we should mention "automatic redeploy on save" is available in one way or another

bfitzpat: Right now we don't handle the automatic redeployment aspect. That would be good to add, though I don't know how we would do that thru the proxy.  

We could do something like if --dev is turned on and they're not using the proxy, we direct output from the deployment to the main Camel-K output channel (and they can always follow the log).

apupier: > Right now we don't handle the automatic redeployment aspect.

I don't understand. If there is the --dev flag, we are supposed to just "delegate" the automatic redeployment to Kamel executable. Nothing to handle on Tooling side.

bfitzpat: It definitely does matter. When we started, we directed ALL camel-k output to a single output channel using the --dev flag. Now we direct nothing from Camel-K directly unless an output channel is explicitly "followed" for the particular deployed integration. So any benefits that --dev provides to the console are lost.

My goal here was consistent operation between using minikube locally with the kamel executable where the --dev flag actually does something and the Rest APIs used across the Kubernetes proxy where there is no --dev equivalent.

bfitzpat: Due to the more explicit steps, we do not automatically redeploy when the file is updated. Again, there's a disconnect between the proxy method and the "kamel" executable method we tried to smooth out to have a consistent operation between the two. There is no way to do automatic redeploy on the proxy side so far as I can tell. So we just have the user re-deploy when a change is made.

apupier: i'm surprised that the automatic redeploy is not workign when using the --dev (but i can't test myself)
So if it is not working and we are currently not using the "follow logs", the --dev options seems to be useless and I think we should completely remove it.

bfitzpat: Automatic redeploy works just fine when using the "kamel" executable with the --dev option at a command line. But we are not re-routing the details from the console output produced by the kamel executable back to the Camel-K output channel. So from the tooling perspective, the --dev option doesn't do anything.

I will add a new issue to work on configuring the --dev option in the tooling so that if it's enabled, we redirect the console output to the Camel-K output channel. But that is beyond the scope of this particular PR.

For the time being, I will remove the --dev options until I come up with a better way to handle it. 

Sound ok @apupier ?

apupier: sounds ok to remove the whole --dev if the option is not working.

bfitzpat:
created https://github.com/camel-tooling/vscode-camelk/issues/56 to handle revised --dev mode functionality

bfitzpat:
@apupier I have removed --dev support for now and will re-introduce with https://github.com/camel-tooling/vscode-camelk/issues/56

To that end, I revised the name of this PR. Hopefully we can close this one and move on :)

Issue #60, opened 2019-08-02
Decorate status of deployed integrations (rest)


bfitzpat:
Follow-up to https://github.com/camel-tooling/vscode-camelk/issues/55

This works for the local kubectl/kamel case but does not yet work for Proxied Rest calls.

bfitzpat:
Ended up being a very minor fix - https://github.com/camel-tooling/vscode-camelk/pull/61

djelinek:
verified - Tooling for Apache Camel K by Red Hat (marketplace version 0.0.9)

PR #61, opened 2019-08-05
issue #60 - get integration status from json returned by rest call

bfitzpat:

### DIFF
--- a/src/CamelKNodeProvider.ts
+++ b/src/CamelKNodeProvider.ts
@@ -192,11 +192,13 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 	// process the JSON we get back from the kube rest API
 	processIntegrationListFromJSON(json : Object) {
 		if (json) {
-			let temp = JSON.stringify(json);
-			let o = JSON.parse(temp);
-			for (var i=0; i<o.items.length;i++) {
-				var integrationName = o.items[i].metadata.name;
-				let newNode = new TreeNode("string", integrationName, 'Running', vscode.TreeItemCollapsibleState.None);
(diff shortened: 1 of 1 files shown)


src/CamelKNodeProvider.ts
```diff
@@ -196,7 +196,8 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 			let o = JSON.parse(temp);
 			for (var i=0; i<o.items.length;i++) {
 				var integrationName = o.items[i].metadata.name;
-				let newNode = new TreeNode("string", integrationName, 'Running', vscode.TreeItemCollapsibleState.None);
+				var status = o.items[i].status.phase;
```
apupier: what do you think to try to provide more explicit naming to `o` and `items[i]`?

I understood that after looking to the context but i'm not even sure.
`o` is the list of integrations represented as JSon?
`o.items[i]` is the Json for one integration?

apupier: I know it was already a problem before but when i read this code again, i really feel that it is so weak. We have no type-safety and neither automated test. If what is returned is modified we will not detect it.

apupier: created https://github.com/camel-tooling/vscode-camelk/issues/63 and https://github.com/camel-tooling/vscode-camelk/issues/62

bfitzpat: Improved

bfitzpat: We have no way to do integration testing at this time. Maybe we can mock it?

apupier: mocking will allow to test our code logic but not to detect that the returned values have changed.

Issue #53, opened 2019-07-29
Handle deployment of integration with configuration via ConfigMap or Secret


bfitzpat:
See https://camel.apache.org/staging/camel-k/latest/configuration/configmap-secret.html

bfitzpat:
In order to provide a config-map, the user will have to have created it ahead of time and set it in `Kubernetes using "kubectl create configmap <name> --from-file=<file.properties>`

I'm thinking we can offer a utility method that takes an application.properties file and generates a config-map yaml file (see https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/ and https://kubectl.docs.kubernetes.io/pages/app_management/secrets_and_configmaps.html)

It appears we can create config-maps with the first call I noted and then swap out "configmap" for "secret" to generate a secret file. 

Now, I have NO clue if we can do any of this generation via Kubernetes Rest APIs, but at least we can get them working on the kubectl/kamel side of things.

bfitzpat:
https://camel.apache.org/staging/camel-k/latest/configuration/configmap-secret.html

bfitzpat:
General flow...
* User creates a config-map or secret yaml file using the kubectl command-line tool
* User right-clicks on an integration and selects "Start Integration with Secret" or "Start Integration with Config-Map", which prompts the user to select the yaml file they want to use.

Behind the scenes we then deploy using `kamel run --secret=[mysecret.file] [integration.file]` or `kamel run --configmap=[myconfigmap.file] [integration.file]`

bfitzpat:
Another good resource - https://matthewpalmer.net/kubernetes-app-developer/articles/ultimate-configmap-guide-kubernetes.html

bfitzpat:
> It appears we can create config-maps with the first call I noted and then swap out "configmap" for "secret" to generate a secret file.

Not quite - turns out there's a slightly different syntax:

`kubectl create secret generic <name> --from-file=<secret.file>`

bfitzpat:
If I create the secret config-map yaml file and deploy it with `kubectl apply -f my-secret.yaml` as per the documentation at https://camel.apache.org/staging/camel-k/latest/configuration/configmap-secret.html, I get it to work. 

If I create the secret using `kubectl create secret generic ...` I get all sorts of weird behavior. 

Created https://github.com/apache/camel-k/issues/900 to try and get some clarity.

bfitzpat:
[...]

PR #64, opened 2019-08-07
issue #53 - adding config-map support - adding secret support - using…

bfitzpat:
… k8s api


apupier:
compilation error reported:

```
src/ConfigMapUtils.ts:75:5 - error TS2322: Type '{}' is not assignable to type 'boolean'.
 75     return new Promise( async (resolve, reject) => {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 76         const kubectl = await k8s.extension.kubectl.v1;
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
 85         }
    ~~~~~~~~~
 86     });
    ~~~~~~~
```

bfitzpat:
> compilation error reported:

Yes, I am aware. That's why this is "work in progress"

bfitzpat:
New menus added for creating config-maps and secrets in kubernetes using the kubectl api. 
image

New drop-down lists added when starting a new integration. Two happen one after the other. First, the user is presented with a list of published config-maps available in the system. 
image

If they want to use one, they select it with the mouse or keyboard and hit enter. If not, they leave the field blank.

Next up they have a drop-down list of published secrets. Again, if they want to use one they select it and move on. If not, they leave it blank.

image

bfitzpat:
My concern is that to publish a simple integration without needing a config-map or secret now requires hitting additional keys or making additional selections. The "flow" is a bit erratic to be reminded of config-maps and secrets each time you start a new integration.

But if we add more menus, that becomes confusing as well. Do we then have three menus -- Start Integration, Start Integration with Config Map, Start Integration with Secret?

What happens when we then add additional resources and individual properties? That becomes two more menus...

We could possibly go with Start Integration and provide a list of options -- Simple, Properties, Resources, Config Map, and Secret -- they choose one and we go down the rabbit hole with their selection. That reduces it to a single click for a simple integration.

bfitzpat:
Changing the flow slightly. Now we front the flow by asking the user what they are trying to accomplish (simple integration, integration with config-map, or integration with secret). Based on their selection, we route them accordingly.

image

A simple integration does not require any further input from the user, so it just does the deed.

Selecting Config Map moves on to enable them to select a config-map from the list of config-maps already registered in the Kubernetes system.

image

Same thing for Secrets.

image

I think this gives us the room we need to move forward with adding Property- and Resource- style deployments in additional PRs.

apupier:
> New menus added for creating config-maps and secrets in kubernetes using the kubectl api.
> image

I think we should not put the Kubernetes action at the top of menu. It won't be the most common actions compared to classic "open" or "compare". I think it would be best to have a separated category and put it close to the bottom.

src/extension.ts
```diff
@@ -201,6 +203,9 @@ export function activate(context: vscode.ExtensionContext) {
 	let startIntegration = vscode.commands.registerCommand('camelk.startintegration', async (uri:vscode.Uri) => { await runTheFile(uri);});
 	context.subscriptions.push(startIntegration);
 
+	// add commands to create config-map and secret objects from .properties files
```
apupier: not sure this comment useful.
renaming the class and the method should be enough to make it clear.

apupier:
I manage to install Kubernetes locally.
The Kubernetes extension can display the perspective. But when i click on "Create Kubernetes Config Map from File", I keep seeing the message: `Kubernetes extensions still coming up, please wait a moment and try again.`

I don't have access to any other errors. By putting breakpoints, I noticed that it is because it doesn't find the "camel.apache.org/v1alpha1" when calling 'api-versions'. I think the error should be clearer for a normal user. (it is normal to not find it as I didn't install it)

src/ConfigMaps.ts
```diff
@@ -0,0 +1,231 @@
...
+			} else {
+				reject("No name given for secret.");
+			}
+		} else {
+			vscode.window.showInformationMessage(`Kubernetes extensions still coming up, please wait a moment and try again.`);
+		}
+	}); 
+}
+
+async function isKubernetesAvailable(): Promise<any> {
```
apupier: I think the method is nto checking if Kubernetes is available but if Camel is available inside Kubernetes (which requires Kubernetes to be available)

apupier:
there is an error when trying to create a config map with space:

`Problem encountered creating new config-map named "myConfigMap with space": Error: error: exactly one NAME is required, got 4 See 'kubectl create configmap -h' for help and examples`

- if not supported by kubectl, needs validation in UI
- if possible, needs to escape/enclosed value i guess

apupier:
well, having errors even with a simple name `test`

`Problem encountered creating new config-map named "test": Error: error: exactly one NAME is required, got 2 See 'kubectl create configmap -h' for help and examples`

In fact it is because the path to the file contains a space

`create configmap new-config-map --from-file=c:\Users\Aurelien Pupier\Desktop\camelfiles\application.properties`

bfitzpat:
I think this is a little cleaner as far as the drop-down that appears after you select "Start Integration"

image

src/ConfigMaps.ts
```diff
@@ -0,0 +1,231 @@
...
+			reject('Kubernetes not available');
+		}
+	}); 
+}
+
+async function getNamedList( itemType : string): Promise<any> {
+	return new Promise( async (resolve, reject) => {
+		const kubectl = await k8s.extension.kubectl.v1;
+		if (kubectl.available) {
+			const sr = await kubectl.api.invokeCommand(`get ${itemType}`);
```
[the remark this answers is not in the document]
bfitzpat: Yes, will change to 'result' to make more clear

src/IntegrationUtils.ts
```diff
@@ -0,0 +1,84 @@
...
+import * as path from 'path';
+import * as child_process from 'child_process';
+import { getConfigMaps, getSecrets } from './ConfigMaps';
+
+export async function startIntegration(context: vscode.Uri): Promise<any> {
+	return new Promise <any> ( async (resolve, reject) => {
+		let selectedConfigMap = undefined;
+		let selectedSecret = undefined;
+		await vscode.window.showQuickPick(['Simple', 'Config-Map', 'Secret'], {
+				placeHolder: 'If your integration requires a Config-map or Secret, select that. Or use Simple for a basic integration.'
```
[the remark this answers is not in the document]
bfitzpat: 'placeHolder' is part of the vscode API - https://code.visualstudio.com/api/references/vscode-api#QuickPickOptions

src/ConfigMaps.ts
bfitzpat: Will spin the camel check into another utility method somewhere else and just rely on kubectl.available for this one.

src/extension.ts
bfitzpat: Leaving the comment. I think it's provides additional context without a lot of overhead (a single comment line).

package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "vscode-camelk",
-	"displayName": "Tooling for Apache Camel-K",
-	"description": "VS Code extensions to support Camel-K functionality",
+	"displayName": "Tooling for Apache Camel K",
+	"description": "VS Code extensions to support Apache Camel K functionality",
```
lhein: extensions? Its just this single one, isn't it?

lhein:
> > New menus added for creating config-maps and secrets in kubernetes using the kubectl api.
> > image
> 
> I think we should not put the Kubernetes action at the top of menu. It won't be the most common actions compared to classic "open". But not sure where is the best place.
> 
> Another thing which seems more important is to have a different category to not mix up with "Open/reveal" actions. But maybe best to handle it in another PR as it applies also for other Camel-K contextual menu command.

see https://github.com/microsoft/vscode/issues/9827
That feature request is open since 2016. No Sub Menus possible. Wondering if its possible to move those commands down the list a bit maybe in an own section bordered by ------------- up and down.

src/ConfigMapAndSecrets.ts
```diff
@@ -0,0 +1,231 @@
...
+			} else {
+				reject("No name given for secret.");
+			}
+		} else {
+			vscode.window.showInformationMessage(`Kubernetes extensions still coming up, please wait a moment and try again.`);
+		}
+	}); 
+}
+
+async function isKubernetesAvailable(): Promise<any> {
```
lhein: You are returning a boolean value always. Wouldn't it be better to specify Boolean instead of Any?

src/ConfigMapAndSecrets.ts
```diff
@@ -0,0 +1,231 @@
...
+	return new Promise( async (resolve, reject) => {
+		const kubectl = await k8s.extension.kubectl.v1;
+		if (kubectl.available) {
+			resolve(true);
+		}
+		reject(false);
```
lhein: Why do u reject here? Isn't reject meant to be used for the error case? resolve(false) seems to be the better way.

src/IntegrationUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+import * as utils from './CamelKJSONUtils';
+import * as path from 'path';
+import * as child_process from 'child_process';
+import { getConfigMaps, getSecrets } from './ConfigMapAndSecrets';
+import * as k8s from 'vscode-kubernetes-tools-api';
+
+export async function startIntegration(context: vscode.Uri): Promise<any> {
+	return new Promise <any> ( async (resolve, reject) => {
+		await vscode.window.showQuickPick([
+			'Basic - Apache Camel K Integration (no ConfigMap or Secret)',
```
lhein: I would opt for creating a constant for these 3 values because they are later on used in a switch too. Given that the values we have here are not short simple words the duplication might easily break the switch one the quick pick values are modified. What do you think?

src/IntegrationUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+			commandString += ` --configmap=${configmap}`;
+		}
+		if (secret && secret.trim().length > 0) {
+			commandString += ` --secret=${secret}`;
+		}
+		commandString += ' "' + absoluteRoot + '"';
+		console.log(`commandString = ${commandString}`);
+		let runKubectl = child_process.exec(commandString, { cwd : foldername});
+		runKubectl.stdout.on('data', function (data) {
+			resolve(true);
```
lhein: you are just checking for any content on stdout. Is that really a sufficient test to determine if the integration has been created successfully?

src/IntegrationUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+		});
+		runKubectl.stderr.on('data', function (data) {
+			utils.shareMessage(extension.mainOutputChannel, `Error deploying ${integrationName}: ${data}`);
+			reject(false);
+			return;
+		});
+	});
+}		
+
+export async function isCamelKAvailable(): Promise<any> {
```
lhein: I would recommend to replace any with boolean because you always return a boolean value.

bfitzpat:
> see [microsoft/vscode#9827](https://github.com/microsoft/vscode/issues/9827)
> That feature request is open since 2016. No Sub Menus possible. Wondering if its possible to move those commands down the list a bit maybe in an own section bordered by ------------- up and down.

I have moved them down like I did the Start Apache Camel-K Integration menu:

image

package.json
bfitzpat: Will change to "VS Code support for Apache Camel K functionality"

src/IntegrationUtils.ts
bfitzpat: Yes, because deploying a new integration takes a long time. This is why we now have decorators and a status tooltip for each "deployed" integration that shows up in the system. The state changes over time through a few states and it's better to get it out there and wait for the result.

bfitzpat:
@lhein thanks to your help with cleaning up the tests and promise issues, we're down to just a couple of things that ONLY crop up on Travis:

```
  ensure that the upstream kubernetes.go sanitize in camel-k have not changed since we checked last
rejected promise not handled within 1 second: undefined
/tmp/sanitize.go
curl -o sanitize.go -L https://raw.githubusercontent.com/apache/camel-k/master/pkg/util/kubernetes/sanitize.go
      ✓ test to see if the sanitize.go file has changed since we stashed it: 429ms
  Suite duration: 1.061 s, Tests: 1
```

and then at the end...

```
rejected promise not handled within 1 second: TypeError: Cannot read property 'stack' of undefined
stack trace: TypeError: Cannot read property 'stack' of undefined
    at t.onMessage.process.on.t.catch.t (/home/travis/build/camel-tooling/vscode-camelk/.vscode-test/vscode-1.37.0/VSCode-linux-x64/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:775:578)
```

I'm not seeing anything obvious on the sanitize.go test, but would be happy with the improved stability we have now and could spin that off into https://github.com/camel-tooling/vscode-camelk/issues/69

bfitzpat:
Took care of one more promise rejected message that was showing up at runtime in the console. Woot! Still have two left in the tests, but we're getting there @lhein

Issue #67, opened 2019-08-09
Avoid refreshing Camel-K integrations view on startup if view is collapsed


apupier:
I noticed it because I don't have "kamel" and the error
```
Refreshing Camel-K Integrations view using kubectl failed. Error: 'kamel' is not recognized as an internal or external command,
operable program or batch file. 
```
is logged.

but it seems reasonable to never do it when the view is collapsed and only call it when the view is expanded.

bfitzpat:
Fixed with https://github.com/camel-tooling/vscode-camelk/pull/83

djelinek:
verified - Tooling for Apache Camel K by Red Hat (marketplace version 0.0.9)

PR #66, opened 2019-08-08
issue #65 - adding extension dependency for k8s extension

bfitzpat:
Signed-off-by: Brian Fitzpatrick <bfitzpat@redhat.com>

PR #83, opened 2019-08-23
issue #67 - avoid refreshing integrations view if collapsed

bfitzpat:


bfitzpat:
Many of the same changes were made in https://github.com/camel-tooling/vscode-camelk/pull/66 - so we should rebase after that gets merged. The main changes are in extension.ts and a utility method in CamelKJSONUtils.ts

Issue #52, opened 2019-07-29
Handle deployment of integration with explicit resources using --resource flag


bfitzpat:
See https://github.com/apache/camel-k/blob/master/examples/resources-route.groovy

"kamel run --resource examples/resources-data.txt examples/resources-route.groovy"

Note that this option does not seem to be documented currently - so need to check with the runtime team on specifics.

PR #85, opened 2019-08-27
issue #52 - adding --resource integration capability

bfitzpat:

### DIFF
--- a/src/CamelKNodeProvider.ts
+++ b/src/CamelKNodeProvider.ts
@@ -100,9 +100,12 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 								this.processIntegrationList(output);
 							}).catch((error) => { 
 								let errMsg : string = error;
-								if (errMsg.toLowerCase().trim().startsWith('error:')) {
-									utils.shareMessage(extension.mainOutputChannel, `Refreshing Apache Camel K Integrations view using kubectl failed. ${error}`);
-									inaccessible = true;
+								if (errMsg) {
+									var errLower =  errMsg.toLowerCase();
+									if (errLower.trimLeft().startsWith('error:')) {
+										utils.shareMessage(extension.mainOutputChannel, `Refreshing Apache Camel K Integrations view using kubectl failed. ${error}`);
+										inaccessible = true;
+									}
 								}
 								reject();
 								return;
--- a/src/IntegrationUtils.ts
+++ b/src/IntegrationUtils.ts
@@ -28,12 +28,23 @@ const devModeIntegration : string = 'Dev Mode - Apache Camel K Integration in De
 const basicIntegration : string = 'Basic - Apache Camel K Integration (no ConfigMap or Secret)';
 const configMapIntegration : string = 'ConfigMap - Apache Camel K Integration with Kubernetes ConfigMap';
 const secretIntegration : string = 'Secret - Apache Camel K Integration with Kubernetes Secret';
+const resourceIntegration : string = 'Resource - Apache Camel K Integration with Resource file';
+
+const ResourceOptions: vscode.OpenDialogOptions = {
+	canSelectMany: true,
+	openLabel: 'Open Resource File(s)',
+	filters: {
+		'Text files': ['txt'],
+		'All files': ['*']
+	}
+};
 
 const choiceList = [ 
 	devModeIntegration,
 	basicIntegration, 
 	configMapIntegration, 
-	secretIntegration ];
+	secretIntegration,
+	resourceIntegration ];
 
 export async function startIntegration(context: vscode.Uri): Promise<any> {
 	return new Promise <any> ( async (resolve, reject) => {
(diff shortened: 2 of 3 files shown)


src/IntegrationUtils.ts
```diff
@@ -27,15 +27,26 @@ import * as k8s from 'vscode-kubernetes-tools-api';
 const basicIntegration : string = 'Basic - Apache Camel K Integration (no ConfigMap or Secret)';
 const configMapIntegration : string = 'ConfigMap - Apache Camel K Integration with Kubernetes ConfigMap';
 const secretIntegration : string = 'Secret - Apache Camel K Integration with Kubernetes Secret';
+const resourceIntegration : string = 'Resource - Apache Camel K Integration with Resource file'
```
apupier: missing semicolon reported by tslint

src/CamelKNodeProvider.ts
```diff
@@ -241,22 +244,20 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
...
 					console.log("[OUT] " + data);
 					shellOutput += data;
 				});
-			}
-			if (runKamel.stderr) {
 				runKamel.stderr.on('data', function (data) {
 					console.log("[ERROR] " + data);
 					reject(data.toString());
 				});
+				runKamel.on("close", () => {
```
apupier: can you explain the advantage of including the 3 actions in the same if branch?
Looks alike we are loosing the case one of the two stout or stderr still available that wants to log something. And also, is the "close" still logged? stdout and stderr are still available when the event is triggered?

src/CamelKNodeProvider.ts
```diff
@@ -100,9 +100,12 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 								this.processIntegrationList(output);
 							}).catch((error) => { 
 								let errMsg : string = error;
-								if (errMsg.toLowerCase().trim().startsWith('error:')) {
-									utils.shareMessage(extension.mainOutputChannel, `Refreshing Apache Camel K Integrations view using kubectl failed. ${error}`);
-									inaccessible = true;
+								if (errMsg) {
+									var errLower =  String.prototype.toLowerCase.apply(errMsg);
```
apupier: why not directly errMsg.toLowerCase() instead of String.prototype.toLowerCase.apply(errMsg) ?

src/IntegrationUtils.ts
```diff
@@ -98,6 +117,11 @@ function createNewIntegration(integrationFileUri: vscode.Uri, configmap? : strin
 		if (secret && secret.trim().length > 0) {
 			commandString += ` --secret=${secret}`;
 		}
+		if (resource) {
+			// assuming that the file is in a path relative to the integration file
```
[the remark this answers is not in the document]
bfitzpat: I agree that this will not be a common use case. Perhaps we move it to being a command only available in the command palette? As for file path, every example I have seen has been relative. We can try an absolute path, but I'm not sure if it will work or not.

src/CamelKNodeProvider.ts
bfitzpat: I believe initially I wasn't declaring errMsg explicitly as a string, so I had to play a bit. I'll see if I can simplify.

src/CamelKNodeProvider.ts
bfitzpat: I honestly can't recall why I shifted it to a single branch at this point. I'll shift it back and see what happens.

src/CamelKNodeProvider.ts
bfitzpat: Looking at this again, it has nothing to do with this PR - so I'll spin this and the other comment off into a separate issue to investigate.

src/CamelKNodeProvider.ts
bfitzpat: Looking at this again, it has nothing to do with this PR - so I'll spin this and the other comment off into a separate issue to investigate.

src/IntegrationUtils.ts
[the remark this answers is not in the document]
bfitzpat: It looks like this will work - `kamel run --resource="/home/bfitzpat/Documents/camelk-examples/resources-data.txt" "resources-route.groovy"`

And "path.normalize" based on the docs will work on the Windows side as well.

src/CamelKNodeProvider.ts
bfitzpat: https://github.com/camel-tooling/vscode-camelk/issues/102

bfitzpat:
I also suspect that we should allow for multiple resources to be specified, so we should be able to apply a similar approach to what I did here for dependencies - https://github.com/camel-tooling/vscode-camelk/pull/93

apupier:
> Questions include... should we support multiple file selection?

> I also suspect that we should allow for multiple resources to be specified, so we should be able to apply a similar approach to what I did here for dependencies - #93

created https://github.com/camel-tooling/vscode-camelk/issues/103

bfitzpat:
Question about deploying integrations with resources... this seems to error out for me for some reason using the resources-route.groovy and resources-data.txt example. 

`kamel run --resource="/home/bfitzpat/Documents/camelk-examples/resources-data.txt" "resources-route.groovy"`

The integration finishes in error state, but I'm not seeing any error in the kit log. I have asked in the Camel K chat where I might look for whatever error occurred.

The documentation in the example uses a relative path, but I'd rather not rely on relative pathing if we can help it.

`kamel run --resource examples/resources-data.txt examples/resources-route.groovy`

Perhaps it's a matter of the quotes?

Issue #54, opened 2019-07-29
Handle deployment of integration with additional properties


bfitzpat:
See https://camel.apache.org/staging/camel-k/latest/configuration/configuration.html

djelinek:
verified - Tooling for Apache Camel K by Red Hat (marketplace version 0.0.9)

Issue #69, opened 2019-08-14
Spend time hunting down errant promise problems


bfitzpat:
While reviewing https://github.com/camel-tooling/vscode-camelk/pull/64, @lhein hit upon several promise issues where we weren't rejecting things accordingly. We need to take a deeper look to ensure there aren't others we have missed.

bfitzpat:
With the improvements in PR https://github.com/camel-tooling/vscode-camelk/pull/64 we are down to:
```
  ensure that the upstream kubernetes.go sanitize in camel-k have not changed since we checked last
rejected promise not handled within 1 second: undefined
/tmp/sanitize.go
curl -o sanitize.go -L https://raw.githubusercontent.com/apache/camel-k/master/pkg/util/kubernetes/sanitize.go
      ✓ test to see if the sanitize.go file has changed since we stashed it: 429ms
  Suite duration: 1.061 s, Tests: 1
```

and then at the end...

```
rejected promise not handled within 1 second: TypeError: Cannot read property 'stack' of undefined
stack trace: TypeError: Cannot read property 'stack' of undefined
    at t.onMessage.process.on.t.catch.t (/home/travis/build/camel-tooling/vscode-camelk/.vscode-test/vscode-1.37.0/VSCode-linux-x64/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:775:578)
```

apupier:
this has been fixed recently by Lars

PR #88, opened 2019-08-28
issue #54 - adding support for deploying with a property

bfitzpat:

### DIFF
--- a/src/IntegrationUtils.ts
+++ b/src/IntegrationUtils.ts
@@ -24,11 +24,14 @@ import * as child_process from 'child_process';
 import { getConfigMaps, getSecrets } from './ConfigMapAndSecrets';
 import * as k8s from 'vscode-kubernetes-tools-api';
 
-const devModeIntegration : string = 'Dev Mode - Apache Camel K Integration in Dev Mode'
+const validNameRegex = /^[A-Za-z][A-Za-z0-9\-\.]*(?:[A-Za-z0-9]$){1}/;
+
+const devModeIntegration : string = 'Dev Mode - Apache Camel K Integration in Dev Mode';
 const basicIntegration : string = 'Basic - Apache Camel K Integration (no ConfigMap or Secret)';
 const configMapIntegration : string = 'ConfigMap - Apache Camel K Integration with Kubernetes ConfigMap';
 const secretIntegration : string = 'Secret - Apache Camel K Integration with Kubernetes Secret';
 const resourceIntegration : string = 'Resource - Apache Camel K Integration with Resource file';
+const propertyIntegration : string = 'Property - Apache Camel K Integration with Property';
 
 const ResourceOptions: vscode.OpenDialogOptions = {
 	canSelectMany: true,
@@ -44,10 +47,12 @@ const choiceList = [
 	basicIntegration, 
 	configMapIntegration, 
 	secretIntegration,
-	resourceIntegration ];
+	resourceIntegration,
+	propertyIntegration
+ ];
 
-export async function startIntegration(context: vscode.Uri): Promise<any> {
-	return new Promise <any> ( async (resolve, reject) => {
+ export async function startIntegration(context: vscode.Uri): Promise<boolean> {
+	return new Promise <boolean> ( async (resolve, reject) => {
 		const choice : string | undefined = await vscode.window.showQuickPick(choiceList, {
 			placeHolder: 'Select the type of Apache Camel K Integration'
 		});
@@ -58,6 +63,7 @@ export async function startIntegration(context: vscode.Uri): Promise<any> {
 			let devMode : boolean = false;
 			let selectedResource : any = undefined;
 			let errorEncountered : boolean = false;
+			let selectedProperty : any = undefined;
 
 			switch (choice) {
 				case devModeIntegration:
@@ -69,12 +75,12 @@ export async function startIntegration(context: vscode.Uri): Promise<any> {
 						if (selectedConfigMap === undefined) {
 							reject (new Error('No ConfigMap selected.'));
 							errorEncountered = true;
-							return;
+							return false;
 						}
 					}).catch ( (error) => {
 						reject(error);
 						errorEncountered = true;
-						return;
+						return false;
 					});
 					break;
 				case secretIntegration:
@@ -83,12 +89,12 @@ export async function startIntegration(context: vscode.Uri): Promise<any> {
 						if (selectedSecret === undefined) {
 							reject (new Error('No Secret selected.'));
 							errorEncountered = true;
-							return;
+							return false;
 						}
 					}).catch ( (error) => {
 						reject(error);
 						errorEncountered = true;
-						return;
+						return false;
 					});
 					break;
 				case resourceIntegration:
@@ -97,12 +103,26 @@ export async function startIntegration(context: vscode.Uri): Promise<any> {
 						if (selectedResource === undefined) {
 							reject (new Error('No Resource selected.'));
 							errorEncountered = true;
-							return;
+							return false;
 						}
 					}).catch ( (error) => {
 						reject(error);
 						errorEncountered = true;
-						return;
+						return false;
+					});
+					break;
+				case propertyIntegration:
+					await getSelectedProperties().then ( (selection) => {
+						selectedProperty = selection;
+						if (selectedProperty === undefined) {
+							reject (new Error('No Property defined.'));
+							errorEncountered = true;
+							return false;
+						}
+					}).catch ( (error) => {
+						reject(error);
+						errorEncountered = true;
+						return false;
 					});
 					break;
 				case basicIntegration:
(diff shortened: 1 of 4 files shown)


bfitzpat:
> I see no big value when providing a single property but that's fine for a first iteration.
I don't disagree. I have added the ability to do multiple properties now using a similar approach to what I did for multiple dependencies.

It results in command lines like `kamel run --property=my.message="test-props" --property=ooga="booga" "props.js"`

But I'm curious what you would recommend for "a test checking the resulting command line would be useful" -- are you thinking we parse the command line to ensure that it would be valid? 

As for an integration test, we don't currently have a way to pull that off. We're not running a Camel K kubernetes somewhere to ping, so unless we use some sort of mock approach that's going to be extremely difficult.

apupier:
> But I'm curious what you would recommend for "a test checking the resulting command line would be useful" -- are you thinking we parse the command line to ensure that it would be valid?

yes, it allows to check that there isn't a bad interaction with other options/branches in our code.

> As for an integration test, we don't currently have a way to pull that off. We're not running a Camel K kubernetes somewhere to ping, so unless we use some sort of mock approach that's going to be extremely difficult.

I know. I mentioned it to pointing out that prioritizing a way to have integration tests is more and more important as we are expanding our usage of the Kamel Cli API.

src/IntegrationUtils.ts
```diff
@@ -52,6 +56,42 @@ export async function startIntegration(context: vscode.Uri): Promise<any> {
...
+							return;
+						} else {
+							const answer = await vscode.window.showQuickPick( ['No', 'Yes'], {
+								placeHolder: 'Are there more properties?',
+								canPickMany : false	});
+							if (!answer) {
+								return;
+							}
+							selectedProperty += `${newProperty}`;
+							let trueAnswer : string = answer as string;
```
apupier: tslint is reporting an unecessary cast

src/IntegrationUtils.ts
```diff
@@ -52,6 +56,42 @@ export async function startIntegration(context: vscode.Uri): Promise<any> {
 						});
 					});
 					break;
+				case propertyIntegration:
+					let hasMoreProperties: boolean = true;
```
apupier: what do you think to extract the propertyIntegration case inside a method, it will make it easier to read all the cases.

src/CamelKNodeProvider.ts
```diff
@@ -99,7 +99,7 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 							await this.getIntegrationsFromCamelK().then((output) => {
 								this.processIntegrationList(output);
 							}).catch((error) => { 
-								let errMsg : string = error;
+								let errMsg : string = error as string;
```
bfitzpat: So this is mostly to avoid the "errMsg.toLowerCase isn't a function" error I see reported sometimes in the Camel K output channel. Though I don't think it's a useful cast, it does remove that problem.

src/extension.ts
```diff
@@ -238,8 +238,8 @@ function removeOutputChannel( channelName: string) {
 	if (outputChannelMap.has(channelName)) {
 		podOutputChannel = outputChannelMap.get(channelName);
 		if (podOutputChannel !== undefined) {
-			podOutputChannel.dispose();
 			outputChannelMap.delete(channelName);
+			podOutputChannel.dispose();
```
bfitzpat: Attempted to change the order to see if it would remove the `Error: Channel has been closed` problem, but it doesn't. It makes sense to remove it and THEN dispose it however, so I left it in place.

src/IntegrationUtils.ts
```diff
@@ -122,8 +140,54 @@ function getSelectedSecret() {
 		});
 }
 
+function getSelectedProperties() {
+	return new Promise <any> ( async (resolve, reject) => {
+		let hasMoreProperties: boolean = true;
+		let returnedProperties : string = '';
```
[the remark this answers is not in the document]
bfitzpat: Good suggestion. Fixed

bfitzpat:
Based on everything I've seen in examples, this code should work fine - but trying to start the props.js example with the property defined in the route as an example doesn't seem to be working for me at the moment. I'm going to reset my minikube instance and see if that helps.

src/IntegrationUtils.ts
```diff
@@ -181,9 +201,52 @@ function getSelectedResource() {
...
+				}).then ( (propValue) => {
+					if (propValue) {
+						propValue = propValue.replace(/"/g, '\\"');
+						newProperty = `${propName}="${propValue}"`;
+					}
+				});
+			});
+			if (!newProperty) {
+				reject(new Error(`No Property provided`));
+				return;
```
lhein: this return looks obsolete

src/IntegrationUtils.ts
```diff
@@ -181,9 +201,52 @@ function getSelectedResource() {
...
+			} else {
+				const answer = await vscode.window.showQuickPick( ['No', 'Yes'], {
+					placeHolder: 'Are there more properties?',
+					canPickMany : false	});
+				if (!answer) {
+					reject(new Error(`No Property answer given`));
+					return;
```
lhein: I am wondering if the code after the reject is even reached.

src/IntegrationUtils.ts
```diff
@@ -181,9 +201,52 @@ function getSelectedResource() {
...
+				}
+				returnedProperties.push(newProperty);
+				if (answer === 'No') {
+					hasMoreProperties = false;
+					resolve(returnedProperties);
+					return;
```
lhein: obsolete return statement

src/IntegrationUtils.ts
```diff
@@ -215,7 +278,11 @@ function createNewIntegration(integrationFileUri: vscode.Uri, devMode? : boolean
 				});
 			}			
 		}
-		commandString += ' "' + absoluteRoot + '"';
+		if (propertyArray && propertyArray.length > 0) {
+			propertyArray.forEach(prop => {
+				commandString += ` -p ${prop}`;
+			});
+		}			
 		console.log(`commandString = ${commandString}`);
```
lhein: do we need to log this?

src/extension.ts
lhein: wondering if the dispose doesn't take care of the delete itself or the other way around and that causes the error message maybe.

src/IntegrationUtils.ts
bfitzpat: Though I understand that it looks that way, I am finding that it doesn't always work that way. For some reason, rejecting doesn't kick out of the method. So I just ensure that it does by putting in the return

src/IntegrationUtils.ts
bfitzpat: See my earlier comment about reject/return

src/IntegrationUtils.ts
bfitzpat: See my earlier comment about reject/return

src/IntegrationUtils.ts
bfitzpat: I find it useful. The user never sees it, but it's been extremely helpful during debugging

src/IntegrationUtils.ts
lhein: Seems that works.

src/extension.ts
bfitzpat: It doesn't seem to. Disposing the object doesn't remove it from the map so far as I can tell.

bfitzpat: Disposing the object doesn't seem to remove it from the map

src/IntegrationUtils.ts
lhein: but there is no more code after this? Just the else case.

src/IntegrationUtils.ts
lhein: there is no more code to execute after that line.

src/extension.ts
lhein: :/

src/IntegrationUtils.ts
bfitzpat: Seems *what* works? reject? reject/return?

src/extension.ts
lhein: see https://github.com/jboss-fuse/vscode-atlasmap/blob/master/src/extension.ts#L139-L186

we have a similar logic there and never experienced issues? Isn't it the same kind of thing?

src/IntegrationUtils.ts
lhein: executing code after the reject

src/IntegrationUtils.ts
bfitzpat: the if handles the case where the user cancels out. the else handles the case where we then check the answer, which they can also cancel out of before we get an answer

src/IntegrationUtils.ts
bfitzpat: I'd rather be consistent than not

src/IntegrationUtils.ts
bfitzpat: Cleaned up

src/IntegrationUtils.ts
bfitzpat: Cleaned up

src/IntegrationUtils.ts
bfitzpat: Cleaned up

src/extension.ts
bfitzpat: will continue to chase this down in https://github.com/camel-tooling/vscode-camelk/issues/69

Issue #87, opened 2019-08-28
Track down "channel has been closed" error


bfitzpat:
Seeing this in the debug console sometimes... 
```
rejected promise not handled within 1 second: Error: Channel has been closed
extensionHostProcess.js:775
stack trace: Error: Channel has been closed
	at l.validate (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:582:372)
	at l.append (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:581:920)
	at l.append (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:675:780)
	at define.createOutputChannel.append.t.then.t (/usr/share/code/resources/app/out/vs/workbench/services/extensions/node/extensionHostProcess.js:583:531)
```

bfitzpat:
https://github.com/camel-tooling/vscode-camelk/pull/89

djelinek:
verified - Tooling for Apache Camel K by Red Hat (marketplace version 0.0.9)

PR #89, opened 2019-08-28
issue #87 - fixing issue with rejected channel promise

bfitzpat:

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -166,16 +166,22 @@ export function activate(context: vscode.ExtensionContext) {
 						if (podName) {
 							let commandString = 'kubectl logs -f ' + podName;
 							let podOutputChannel = getOutputChannel(podName);
-							podOutputChannel.show();
+							if (podOutputChannel) {
+								podOutputChannel.show();
+							}
 							let kubectl = child_process.exec(commandString);
 							if (kubectl.stdout) {
 								kubectl.stdout.on('data', function (data) {
-									podOutputChannel.append(`${data}`);
+									if (podOutputChannel) {
+										utils.shareMessage(podOutputChannel, `${data}`);
+									}
 								});
 							}
 							if (kubectl.stderr) {
 								kubectl.stderr.on('data', function (data) {
-									mainOutputChannel.append("[ERROR] " + `${data} \n`);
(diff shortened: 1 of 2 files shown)


bfitzpat:
Not sure about this one - the issue does go away, but I'm not convinced we're actually disposing of the output channel under the covers.

src/extension.ts
```diff
@@ -230,11 +236,13 @@ export function hideStatusLine() {
...
 function removeOutputChannel( channelName: string) {
 	let podOutputChannel;
-	if (outputChannelMap.has(channelName)) {
+	if (outputChannelMap !== undefined) {
 		podOutputChannel = outputChannelMap.get(channelName);
 		if (podOutputChannel !== undefined) {
-			podOutputChannel.dispose();
 			outputChannelMap.delete(channelName);
+			// concerned that this doesn't actually delete the resources associated with the output channel, 
+			// but dispose causes other problems
```
apupier: what are the other problems with dispose?

bfitzpat: We were disposing the channel before (podOutputChannel.dispose(); on extension.ts line 236 or so) and for some reason there seems to be a thread somewhere still trying to update it in the background AFTER it's been disposed, which was what was creating the "channel has been closed" error I think. Dispose is what SHOULD be called it seems, but doesn't look like it's behaving quite as expected.

apupier: it means that we are not disposing the pod output channel at all now, isn't it? If yes, sounds like a potential memory leak.

bfitzpat: We dispose of all output channels when the extension is deactivated. https://github.com/camel-tooling/vscode-camelk/blob/bce6fa9d91e9518cdb00c57372a6695da01d81ed/src/extension.ts#L473

But yes, other than that we are not disposing of them.

apupier: the extensions not deactivated in a classic usage.
it means that we are trading errors in log against a memory leak. I prefer the errors in log.

bfitzpat: Ok. As the errors in the debug log don't affect the user, we'll just let it be then. Will close this PR

bfitzpat:
Closing this PR per the conversation

Issue #51, opened 2019-07-29
Handle deployment of integration with explicit dependencies using -d or --dependency flag


bfitzpat:
See https://github.com/apache/camel-k#dependencies-and-component-resolution

See https://github.com/apache/camel-k/blob/master/examples/routes-rest.js as an example

i.e. "kamel run --name=withrest --dependency=camel-undertow examples/routes-rest.js"

bfitzpat:
Multiple dependencies seem to be able to be set `kamel run -d mvn:com.google.guava:guava:26.0-jre -d camel-mina2 Integration.java`

bfitzpat:
I have some code working where it can generate: `kamel run --dependency=do-me --dependency=group:artifact:version "routes.js"`

Need to figure out if we have a way to validate camel dependencies or maven dependencies thru node.js

djelinek:
verified - Tooling for Apache Camel K by Red Hat (marketplace version 0.0.9)

PR #93, opened 2019-08-30
issue #51 - adding ability to specify explicit dependencies when depl…

bfitzpat:
…oying integration

### DIFF
--- a/src/IntegrationUtils.ts
+++ b/src/IntegrationUtils.ts
@@ -32,6 +32,7 @@ const configMapIntegration : string = 'ConfigMap - Apache Camel K Integration wi
 const secretIntegration : string = 'Secret - Apache Camel K Integration with Kubernetes Secret';
 const resourceIntegration : string = 'Resource - Apache Camel K Integration with Resource file';
 const propertyIntegration : string = 'Property - Apache Camel K Integration with Property';
+const dependencyIntegration : string = 'Dependencies - Apache Camel K Integration with Explicit Dependencies';
 
 const ResourceOptions: vscode.OpenDialogOptions = {
 	canSelectMany: true,
@@ -48,7 +49,8 @@ const choiceList = [
 	configMapIntegration, 
 	secretIntegration,
 	resourceIntegration,
-	propertyIntegration
+	propertyIntegration,
+	dependencyIntegration
  ];
 
  export async function startIntegration(context: vscode.Uri): Promise<boolean> {
@@ -64,7 +66,8 @@ const choiceList = [
 			let selectedResource : any = undefined;
 			let errorEncountered : boolean = false;
 			let selectedProperty : any = undefined;
-
+			let selectedDependency : any = undefined;
+			
 			switch (choice) {
 				case devModeIntegration:
 					devMode = true;
@@ -125,13 +128,27 @@ const choiceList = [
 						return false;
 					});
 					break;
+				case dependencyIntegration:
+					await getSelectedDependencies().then ( (selection) => {
+						selectedDependency = selection;
+						if (selectedDependency === undefined) {
+							reject (new Error('No Dependencies defined.'));
+							errorEncountered = true;
+							return;
+						}
+					}).catch ( (error) => {
+						reject(error);
+						errorEncountered = true;
+						return;
+					});
+					break;
 				case basicIntegration:
 					// do nothing with config-map or secret
 					break;
 			}
 				
 			if (!errorEncountered) {
-				await createNewIntegration(context, devMode, selectedConfigMap, selectedSecret, selectedResource, selectedProperty)
+				await createNewIntegration(context, devMode, selectedConfigMap, selectedSecret, selectedResource, selectedProperty, selectedDependency)
 					.then( success => {
 						if (!success) {
 							reject(false);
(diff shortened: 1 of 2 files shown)


bfitzpat:
image

image

image

image

image

bfitzpat:
There are two things that need to be expanded on for this functionality.
1. Validating the name of a given camel component. Right now we don't have a good way to do that inside JS/TS
2. Validating that a given maven dependency actually exists and is accessible

@lhein or @apupier - any ideas how to improve on these? 

Right now all I'm doing is validating that there are three colons (group:artifact:version) (though "group:artifact:" is valid even though it shouldn't be)... and validating that there's a dash (i.e. camel-component). This is obviously not enough validation, so I'd love some input on how to improve it

bfitzpat:
On the plus side, we should be able to apply the same multi-prompt logic for other types of deployments such as properties https://github.com/camel-tooling/vscode-camelk/pull/88 or resources https://github.com/camel-tooling/vscode-camelk/pull/85

apupier:
> 2\. Validating that a given maven dependency actually exists and is accessible

might used Maven command line to see if a dependency is available or not: https://www.npmjs.com/package/maven

or use https://www.npmjs.com/package/mvn-artifact-download and try download to see if it is available

subsidiary question: how/which extra repositories are configured for Camel K?

apupier:
> 1. Validating the name of a given camel component. Right now we don't have a good way to do that inside JS/TS

you mean a camel-component supported by Camel K? or that the name is a syntaxically valid component name?

bfitzpat:
> 
> 
> > 1. Validating the name of a given camel component. Right now we don't have a good way to do that inside JS/TS
> 
> you mean a camel-component supported by Camel K? or that the name is a syntactically valid component name?

Ideally both -- is it a valid component name and is it downloadable/accessible

bfitzpat:
> > 2. Validating that a given maven dependency actually exists and is accessible
> might used Maven command line to see if a dependency is available or not: https://www.npmjs.com/package/maven
> or use https://www.npmjs.com/package/mvn-artifact-download and try download to see if it is available

I looked at the NPM Maven component and didn't see a way to use it to test a dependency. Maybe mvn-artifact-dependency could work, but I don't see an easy way to try that. Do you have any ideas?

> subsidiary question: how/which extra repositories are configured for Camel K?

Considering the whole Apache Camel K project is written in Go with very little maven, I can't see anywhere they're explicitly calling out particular repositories - but it does seem that they're getting Camel 3 artifacts from somewhere.

apupier:
> > > 1. Validating that a given maven dependency actually exists and is accessible
> > >    might used Maven command line to see if a dependency is available or not: https://www.npmjs.com/package/maven
> > >    or use https://www.npmjs.com/package/mvn-artifact-download and try download to see if it is available
> 
> I looked at the NPM Maven component and didn't see a way to use it to test a dependency. Maybe mvn-artifact-dependency could work, but I don't see an easy way to try that. Do you have any ideas?

I never tried and don't the API but i would expect that at least, when trying to download an artefact if not available, it is throwing an error that can be caught to determine if the artifact is missing (for big artifacts it can cause a delay).
Hope that there are some smarter way to do it in one for the npm packages, needs investigation

apupier:
> > subsidiary question: how/which extra repositories are configured for Camel K?
> 
> Considering the whole Apache Camel K project is written in Go with very little maven, I can't see anywhere they're explicitly calling out particular repositories - but it does seem that they're getting Camel 3 artifacts from somewhere.

Camel 3 artifacts are available on Central Maven
http://repo1.maven.org/maven2/org/apache/camel/camel-core/3.0.0-RC1/

It is possible to configure extra Maven repositories, https://github.com/apache/camel-k#configure-maven-settings . As they don't have a Project structure,  it requires to allow the maven repo configuration from the same command and to connect to Camel K instance to retrieve the already configured repo. This seems a bit too complicated for a first iteration.

bfitzpat:
@apupier seems we just leave it pretty open for now and do more validation in the future when things are a bit more well defined on this front?

apupier:
> seems we just leave it pretty open for now and do more validation in the future when things are a bit more well defined on this front?

it seems a fair amount of work, so I vote for that too. Can I let you open issue(s) with opened questions/identified pain points?

bfitzpat:
@apupier - spun off a new issue https://github.com/camel-tooling/vscode-camelk/issues/117

src/IntegrationUtils.ts
```diff
@@ -197,7 +246,7 @@ function createNewIntegration(integrationFileUri: vscode.Uri, devMode? : boolean
 				console.error(error);
```
lhein: do we need to log it?

src/IntegrationUtils.ts
```diff
@@ -207,15 +256,18 @@ function createNewIntegration(integrationFileUri: vscode.Uri, devMode? : boolean
...
-			}			
+			});
+		}			
+		if (dependencyArray && dependencyArray.length > 0) {
+			dependencyArray.forEach(dependency => {
+				commandString += ` --dependency=${dependency}`;
+			});
 		}
-		commandString += ' "' + absoluteRoot + '"';
 		console.log(`commandString = ${commandString}`);
```
lhein: logging required?

bfitzpat: Yes, I like having it there more for debugging purposes than anything. The user never sees it.

src/IntegrationUtils.ts
```diff
@@ -257,3 +310,19 @@ export async function isCamelKAvailable(): Promise<boolean> {
...
+function validateDependency(text : string) {
+	let inputEmpty : boolean = (text === undefined || text.trim().length === 0);
+	if (inputEmpty) {
+		return 'Dependency must be specified in the form of an Apache Camel component id or a Maven dependency in the form of group:artifact:version';
+	}
+	if (!inputEmpty && text.indexOf(':') > -1) {
+		let trifold : string[] = text.split(':');
+		// make sure we have three colon-delimited strings
+		return !(trifold && trifold.length === 3) ? 'Dependency must be specified in the form of a Maven dependency as group:artifact:version' : null;
+	} else if (!(!inputEmpty && text.indexOf('-') > -1)) {
```
apupier: if I understand well, it requires a `-` in the component id.
I don't think that it is a requirements. For instance there are ahc, apns, amqp, file, timer, ... Or I'm looking to the wrong thing?

lhein: its about artifactId I thought....like camel-ftp and so on

bfitzpat: Lars is right - if you look here - https://camel.apache.org/components/latest/ - and click through to ahc, apns, etc. - the dependency always seems to boil down to camel-component

src/IntegrationUtils.ts
```diff
@@ -257,3 +310,19 @@ export async function isCamelKAvailable(): Promise<boolean> {
...
+		// assume we're a camel component for now
+		return 'Dependency must be specified in the form of an Apache Camel component id or a Maven dependency in the form of group:artifact:version';
```
lhein: maybe the text is confusing?

src/IntegrationUtils.ts
apupier: ok, then maybe worth precising i tin the message that i tis the "camel component artifact id" and not the "camel component id"

src/IntegrationUtils.ts
bfitzpat: Recommendation? Maybe change to "Apache Camel component Artifact Id"?

bfitzpat: Updated

Issue #106, opened 2019-09-05
[Regression]Quickpick used for Deploy command is flashing and not giving user a chance to provide input


bfitzpat:
This week an error cropped up in VS Code 1.38.0 that causes the integration type dropdown to not function properly. It flashes and steps through, not giving the user an opportunity to make a selection. It was also brought up that some of that code is unnecessarily complex, so it needs to be streamlined in the process.

bfitzpat:
https://github.com/camel-tooling/vscode-camelk/pull/107 fixes the issue

apupier:
I think I found the original reason because quickPick is working in our othe rplugins.
it is due to the outputchannel which is always shown https://github.com/camel-tooling/vscode-camelk/commit/c5bd8217876b8161192b4307795e5a7ec750f1ff#diff-45961516239bc4b63c0fc662708b0e70R148 on each message. Which is making the focus to be lost.

it explains:
- it doesn't disappear anymore which ignoreFocusOut to true
- the up/down arrow not working when ignoreFocusOut set to true

djelinek:
verified - Tooling for Apache Camel K by Red Hat (marketplace version 0.0.9)

PR #107, opened 2019-09-05
issue #106 - restructuring integration menu dropdown code

bfitzpat:

### DIFF
--- a/src/IntegrationUtils.ts
+++ b/src/IntegrationUtils.ts
@@ -28,54 +28,100 @@ const basicIntegration : string = 'Basic - Apache Camel K Integration (no Config
 const configMapIntegration : string = 'ConfigMap - Apache Camel K Integration with Kubernetes ConfigMap';
 const secretIntegration : string = 'Secret - Apache Camel K Integration with Kubernetes Secret';
 
+const choiceList = [ basicIntegration, 
+	configMapIntegration, 
+	secretIntegration ];
+
 export async function startIntegration(context: vscode.Uri): Promise<any> {
 	return new Promise <any> ( async (resolve, reject) => {
-		await vscode.window.showQuickPick([ basicIntegration, configMapIntegration, secretIntegration ],
-		{
+		const choice : string | undefined = await vscode.window.showQuickPick(choiceList, {
 			placeHolder: 'Select the type of Apache Camel K Integration'
-		}).then( async (choice) => {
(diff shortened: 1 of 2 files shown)


bfitzpat:
@lhein I think this is a little easier to understand (and better yet, it works!) -- but am curious if you see some way to streamline the switch statement. That seems clunky to me even though it works.

bfitzpat:
Once this one is in place, I will work on updating https://github.com/camel-tooling/vscode-camelk/pull/85 and https://github.com/camel-tooling/vscode-camelk/pull/88 and https://github.com/camel-tooling/vscode-camelk/pull/91 and https://github.com/camel-tooling/vscode-camelk/pull/93

apupier:
I think the original problem is fixed by https://github.com/camel-tooling/vscode-camelk/pull/110 , this PR is providing a workaround with some limitations.

This code refactor still makes sense as it is providing cleaner code.
The ignoreFocusOut can be used or not, the good side is that it will avoid a too big same kind of regression , we will just have the up/down arrow not working instead of not being able to use the command at all.

bfitzpat:
Arrows work now without the ignoreFocusOut and it's been rebased on the latest master that includes the fix https://github.com/camel-tooling/vscode-camelk/pull/110. @apupier 

If we get this one approved and merged, I'll work on the other PRs to get them refactored in this manner so they are all a little easier to read.

src/IntegrationUtils.ts
```diff
@@ -28,54 +28,100 @@ const basicIntegration : string = 'Basic - Apache Camel K Integration (no Config
...
 
+const choiceList = [ basicIntegration, 
+	configMapIntegration, 
+	secretIntegration ];
+
 export async function startIntegration(context: vscode.Uri): Promise<any> {
 	return new Promise <any> ( async (resolve, reject) => {
-		await vscode.window.showQuickPick([ basicIntegration, configMapIntegration, secretIntegration ],
-		{
+		const choice : unknown = await vscode.window.showQuickPick(choiceList, {
```
apupier: shouldn't we use "string | undefined" instead of "unknown"?

bfitzpat: Considering I had no clue I could do that. :) I just tried it and it worked great. Changing now.

bfitzpat:
Done. Thanks for the help

PR #110, opened 2019-09-06
Avoid showing output on each message #106

apupier:
it introduced a regression to use the Start integration command palette

also, if a user has hidden or closed the output panel, he/she doesn't
want to have the output popping on each information/trace message


bfitzpat:
Good catch.

PR #115, opened 2019-09-10
Add support for starting integrations with YAML file extension

### LINKED ISSUE #112: Add support for starting integrations with YAML file extension
See example here:

https://github.com/apache/camel-k/blob/master/examples/routes.yaml

I could try to take a stab at implementing this.

jamesnetherton:
fixes #112

I didn't find a simple way of implementing an automated way of testing this, so I tested manually against a local cluster and it works nicely.


jamesnetherton:
I should mention that the example YAML file I refer to in #112 doesn't work too well because of an issue in camel-k (See https://github.com/apache/camel-k/issues/929) .

Better to test with [this](https://github.com/apache/camel-k/blob/master/e2e/files/yaml.yaml) one.

bfitzpat:
I would request one change - can we add a note to the CHANGELOG.md file for this?

jamesnetherton:
I added an entry to `CHANGELOG.md`.

Issue #58, opened 2019-08-02
Investigate auto-refresh of integrations view


bfitzpat:
Now that we are adding decorators to the Camel-K Integrations view that update on refresh, it would be good to set an auto-refresh cycle so the user wouldn't have to manually click the Refresh button to see changes. (https://github.com/camel-tooling/vscode-camelk/issues/55#issuecomment-517593459)

djelinek:
verified - Tooling for Apache Camel K by Red Hat (marketplace version 0.0.9)

PR #130, opened 2019-09-13
issue #58 - adding a listener for server changes to published integra…

bfitzpat:
…tions

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -25,6 +25,7 @@ import * as rp from 'request-promise';
 import {platform} from 'os';
 import * as configmapsandsecrets from './ConfigMapAndSecrets';
 import * as integrationutils from './IntegrationUtils';
+import * as events from 'events';
 
 export let mainOutputChannel: vscode.OutputChannel;
 export let myStatusBarItem: vscode.StatusBarItem;
@@ -36,9 +37,10 @@ let curlOption : string = '-c';
 let proxyPort : number;
 let showStatusBar : boolean;
 let camelKIntegrationsTreeView : vscode.TreeView<TreeNode>;
+let eventEmitter = new events.EventEmitter();
+const restartKubectlWatchEvent = 'restartKubectlWatch';
 
 // This extension offers basic integration with Camel K (https://github.com/apache/camel-k) on two fronts.
-
 export function activate(context: vscode.ExtensionContext) {
 
 	outputChannelMap = new Map();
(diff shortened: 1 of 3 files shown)


bfitzpat:
So a couple of things here. I'm using kubectl to monitor the system and react when anything changes with published "integration" objects in the server. I'm using "kubectl get integrations -w"

This works great for the traditional case where everything runs fine. I get notified for most states if they work all the way through Building Kit into Running, but I'm not getting notified when things go from Building Kit to Error. 

Wondering if that's a missed notification being passed along from Camel K back to Kubernetes? Have asked on the Camel K chat but also opened an issue - https://github.com/apache/camel-k/issues/937

bfitzpat:
I can probably get around this watching the integrationkits thread instead - "kubectl get integrationkits -w" - trying that now.

bfitzpat:
Nope - that doesn't work - and eventually the watch thread quits as well, so this may not be the best way to go.

bfitzpat:
If I can find a way to keep the kubectl call from timing out after a certain amount of inactivity, maybe this will work.

lhein:
it feels like this feature is not ready yet then?

bfitzpat:
> it feels like this feature is not ready yet then?
Still working on it. I'm guessing there's a way to work around the kubernetes timeout. And the Building Kit/Error problem is being fixed on the Camel K side already courtesy of @jamesnetherton (see https://github.com/apache/camel-k/pull/927)

bfitzpat:
@lhein Found a workaround - now we simply restart the watch when it times out and all is good with the world. Even did some additional cleanup to remove now unnecessary code in the node provider. We used to do multiple `kamel get` calls to see if we could detect changes, but now that we're using the watch we don't need to do so.

lhein:
ok, is there not a parameter to watch in endless mode?

bfitzpat:
> ok, is there not a parameter to watch in endless mode?

Not that I can find in the kubectl guide - https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#get

And looking online it looks like the timeout is customizable on the server side, so it's nothing I can pass on the cli.

src/extension.ts
```diff
@@ -100,6 +101,15 @@ export function activate(context: vscode.ExtensionContext) {
...
 	});
 
+	// start the watch listener for auto-updates 
+	startListeningForServerChanges();
+
+	// Listener to handle auto-refresh of view - kubectl times out, so we simply restart the watch when it does
+	var watchListener = function restartKubectlListenerOnEvent() {
+		startListeningForServerChanges();
+	};
+	eventEmitter.on('restartKubectlWatch', watchListener);
```
apupier: what about using a constant for 'restartKubectlWatch' so that in case of a refactor we are sure that all places are changed accordingly

Issue #131, opened 2019-09-16
Support mixing --dev mode and other integration types


bfitzpat:
This is related to https://github.com/camel-tooling/vscode-camelk/issues/95 but a bit more specific. I think we want to enable switching on/off a --dev mode option that enables us to use any of the integration types in dev mode more easily. 

Investigate adding a simple command that can be invoked from the command palette to enable --dev mode on top of --property, --resource, --secret, etc.

bfitzpat:
@lhein pointed me to https://github.com/jboss-fuse/vscode-atlasmap/blob/master/src/extension.ts#L192-L205 as a possible fix for the child process "kill" problem, but it doesn't quite work. 

The issue here is that when we start a "kamel integration --dev" process, we're never getting a close/exit signal indicating that the process is finished. So it's keeping it open. This means that if we follow this approach:

* deploy integration in --dev mode
* remove integration in --dev mode
* deploy same integration without --dev mode

We are still seeing the main Apache Camel K output channel get polluted with log entries from the running integration that it STILL thinks is in dev mode because we never killed our original dev mode publishing. 

Removing the integration SHOULD kill the initial --dev mode publishing I think. Or we should get some kind of signal (close/exit) when the --dev mode integration is removed. 

I created https://github.com/apache/camel-k/issues/938 on the kamel cli side, but I'll continue to try and find a way around it in the tooling

bfitzpat:
I think we can consider this closed with the merging of https://github.com/camel-tooling/vscode-camelk/pull/302

PR #132, opened 2019-09-16
issue #131 - adding command palette option to turn dev mode on and off

bfitzpat:

### DIFF
--- a/src/IntegrationUtils.ts
+++ b/src/IntegrationUtils.ts
@@ -53,6 +53,9 @@ const choiceList = [
 	dependencyIntegration
  ];
 
+ let childProcessMap : Map<string, child_process.ChildProcess>;
+
+
  export async function startIntegration(context: vscode.Uri): Promise<boolean> {
 	return new Promise <boolean> ( async (resolve, reject) => {
 		const choice : string | undefined = await vscode.window.showQuickPick(choiceList, {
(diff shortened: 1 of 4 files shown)


bfitzpat:
Essentially the thought here is that the --dev mode option can be turned on temporarily in a given session. So the user goes to the command palette and types "Apache Camel K: Enable --dev mode for all future deployments", works on their integration, and deploys it maybe in --configmap mode. It's essentially --configmap and --dev on the command line that we're invoking.

The user should then be able to turn the mode off by going back to the command palette and typing "Apache Camel K: Disable --dev mode..."

src/IntegrationUtils.ts
```diff
@@ -322,6 +325,8 @@ function createNewIntegration(integrationFileUri: vscode.Uri, devMode? : boolean
 				console.error(error); 
 			});
 
+		devMode = extension.getDevMode();
```
apupier: What is happening when use is using the "Start integration in dev mode" but never "started dev mode" or either "stopped dev mode" previously? it looks alke the command "Start integration in dev mode will not start in dev mode.

bfitzpat: Taking a look

bfitzpat: Updated

src/IntegrationUtils.ts
```diff
@@ -322,41 +325,55 @@ function createNewIntegration(integrationFileUri: vscode.Uri, devMode? : boolean
...
 		}			
-		console.log(`commandString = ${commandString}`);
+		console.log(`commandString = ${kamelCLI} ${kamelArgs}`);
 		if (devMode && devMode === true) {
 			if (extension.mainOutputChannel) {
 				extension.mainOutputChannel.show();
 			}
 		}
-		let runKubectl = child_process.exec(commandString, { cwd : foldername});
+		let runKubectl = child_process.spawn(kamelCLI, kamelArgs, { cwd : foldername});
```
bfitzpat: Using child_process.exec was creating a subprocess that was not getting killed when the main process was killed, so switching to spawn eliminates that issue and we have a much cleaner process that can be killed properly.

bfitzpat: This was the key - https://stackoverflow.com/questions/27770537/two-processes-created-by-child-process-exec

bfitzpat:
One important note. This is only a per-session setting. It does not persist currently from session to session, so if you shut down VS Code and restart, it starts back at "--dev disabled" mode. I'm not sure if we want to make it a setting or not. Any thoughts? @apupier @lhein ?

lhein:
If we want to persist it then we will have to drop the command and introduce a setting. In my opinion that is the way to go. Maybe we should look at other commands if there are any commands that alter settings. If that is the case we can surely keep the command too.

src/IntegrationUtils.ts
```diff
@@ -322,41 +325,55 @@ function createNewIntegration(integrationFileUri: vscode.Uri, devMode? : boolean
 				console.error(error); 
 			});
 
-		let commandString = `kamel run "${absoluteRoot}"`;
+		if (extension.getDevMode() === true) {
+			// it true, always override whatever the dev mode is that's passed in
```
apupier: typo comment it -> if

bfitzpat:
Ok. I will spin the exec => spawn change off into a separate change then -- that needs to get in to avoid the hanging process. But will leave the rest for a future PR.

bfitzpat:
Spun off the exec to spawn work in https://github.com/camel-tooling/vscode-camelk/pull/140 and will continue work on the --dev mode stuff in a later release.

bfitzpat:
Closing this one and will revisit in another PR at another date

PR #133, opened 2019-09-17
extracting logic from activate into separate functions

lhein:
Starting to make code more readable.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -29,6 +29,7 @@ import * as events from 'events';
 
 export let mainOutputChannel: vscode.OutputChannel;
 export let myStatusBarItem: vscode.StatusBarItem;
+
 let camelKIntegrationsProvider = new CamelKNodeProvider();
 let useProxy : boolean;
 let outputChannelMap : Map<string, vscode.OutputChannel>;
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -560,5 +513,72 @@ function startKubeProxy(): Promise<string> {
 				reject(new Error(data.toString()));
 			});
 		}
-	}); 
+	});
 }
+
+function determinePlatform(): void {
```
apupier: The real goal is more to determine the command curl based on the detected platform than to determine platform.

src/extension.ts
```diff
@@ -560,5 +513,72 @@ function startKubeProxy(): Promise<string> {
...
+	const osType = platform();
+	if (osType === 'win32') {
+		curlCommand = 'cmd';
+		curlOption = '/c';
+	}
+}
+
+function applyStatusBarSettings(): void {
+	// process the workspace setting indicating whether we should use the proxy or CLI
```
apupier: the comment is unrelated to the code

src/extension.ts
lhein: true. I will rename it.

src/extension.ts
lhein: that is nothing I added. I am just extracting to functions.

apupier: I would prefer to remove the 3 comment lines that are even more useless as the refactoring is making the code clearer, and allows to easily spot that one of them is incorrect

lhein: done

lhein:
rebased the PR

src/extension.ts
```diff
@@ -589,4 +542,70 @@ function startListeningForServerChanges() {
...
+}
+
+function applyUserSettings(): void {
+	applyStatusBarSettings();
+	applyProxySettings();
+	applyProxyPortSettings();
+}
+
+function createIntegrationsView(): void {
+	// create the integrations view
```
apupier: this comment is really really redundant

lhein: removed

PR #134, opened 2019-09-17
improve handling of promises to prevent unhandled error cases

lhein:


lhein:
broke 2 tests....investigating

lhein:
tests fixed

src/extension.ts
```diff
@@ -560,5 +539,72 @@ function startKubeProxy(): Promise<string> {
 				reject(new Error(data.toString()));
 			});
 		}
-	}); 
+	});
+}
+
+function determinePlatform(): void {
```
apupier: The real goal is more to determine the command curl based on the detected platform than to determine platform.

src/extension.ts
```diff
@@ -560,5 +539,72 @@ function startKubeProxy(): Promise<string> {
...
+	const osType = platform();
+	if (osType === 'win32') {
+		curlCommand = 'cmd';
+		curlOption = '/c';
+	}
+}
+
+function applyStatusBarSettings(): void {
+	// process the workspace setting indicating whether we should use the proxy or CLI
```
apupier: the comment is unrelated to the code

src/CamelKNodeProvider.ts
```diff
@@ -62,7 +62,7 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 			}
 			return Promise.resolve(oldNodes);
 		}
-		return Promise.reject();
+		return Promise.reject(undefined);
```
apupier: what is the advantage of rejecting with undefined compared to rejecting with empty method?

Shouldn't we try to reject with an Error?

apupier:
reviewing https://github.com/camel-tooling/vscode-camelk/pull/133 first is easier

src/extension.ts
lhein: see other PR comment

src/extension.ts
lhein: see other PR comment

src/CamelKNodeProvider.ts
lhein: Yes, rejecting with an error is probably best. Though it is actually making no difference in the result. Rejects will be handled in the catch block and resolves will be handled in the then block. If the catch block is not taking care of the error then it is not important what you put into the reject. From what I see in the comment that function is only used in the tests, but I haven't double checked yet.

lhein: Checked and its used internal in the CamelKNodeProvider and in a test. The positions inside CamelKNodeProvider where it is used don't look at the reject cause. I think that whole class needs some refactor at some point but that was not the subject for this PR. If you are fine with it I would like to defer it to a separate PR just refactoring this file.

apupier: so maybe we can keep promise.reject() instead of introducing Promise.reject(undefined) which is bringing nothing more (or I misunderstood).
And ok to create another issue to improve this part with better error handling

lhein: I am fine with that. Will revert to empty reject in both positions.

lhein: Now I see why I added the parameter. Code complains that it should have a parameter. Seems to come from the linter. Anyways it builds and runs fine, so I will keep it without parameter for now until we refactor that file.

lhein:
rebased the PR

src/test/camelkjsonutils.test.ts
```diff
@@ -48,7 +48,11 @@ suite('ensure camelk utilities work as expected', function() {
 					assert.ok(output.length > 0);
 					done();
 				});
-		});
+		})
+		.catch((err) => {
```
apupier: why do we need to put the catch here although there are not in the following test "should be able to construct usable rest URL" ?

(I really don't get how Typescript is handling their exception mechanism)

lhein: The catch is for the stringify call because it can be rejected. In the followed test I don't see a stringify? Or do I miss something?

apupier: no, there is no stringify.
You are nto missing something, I just don't understand the philosophy of this language :-)

lhein: As far as I understood it a Promise is handled that if you resolve it, it will return the the .then() section of the calling function. If the promise gets rejected it will end up in the .catch() block of the calling function. I am not saying I am fully understanding that language and I might be wrong too. :D  Have a look at https://basarat.gitbooks.io/typescript/docs/promise.html

PR #136, opened 2019-09-19
refactored the CamelKNodeProvider

lhein:

### DIFF
--- a/src/CamelKNodeProvider.ts
+++ b/src/CamelKNodeProvider.ts
@@ -196,7 +196,7 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 	getIntegrationsFromCamelKRest(): Promise<Object> {
 		return new Promise( (resolve, reject) => {
 			let proxyURL = utils.createCamelKRestURL();
-			const options = {
+			let options = {
 				uri: proxyURL,
 				headers: {
 					'Content-Type': 'application/json',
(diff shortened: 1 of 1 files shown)


src/CamelKNodeProvider.ts
```diff
@@ -49,92 +49,85 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 	}
 
 	// get the list of children for the node provider
-	public getChildren(element?: TreeNode): Thenable<TreeNode[]> {
+	public getChildren(element?: TreeNode): Promise<TreeNode[]> {
```
apupier: what is the advantage of changing from Thenable to promise?

lhein: It returns a Promise and I see no benefit in using Thenable here. So why should we obscure the real return type?

apupier: I don't know why Thenable was used initially here.
Thenable seems more generic than Promise.
I have no idea what is the best practice, asking to try to understand it.

lhein: best I could find atm is https://stackoverflow.com/questions/29435262/regarding-promises-a-specification-what-is-the-difference-between-the-terms-t

I think you mistake it for the "better return the interface" paradigm in Java which I think doesn't fit here.

lhein: But to be perfectly honest I am not able to give you a 100% answer here. I am myself too new to JS/TS.

apupier: still don't understand what is best to use but at least we are more "coherent" with promise as it is what we are using most of the time in our codebase

bfitzpat:
Closed with merging of combined PR https://github.com/camel-tooling/vscode-camelk/pull/141

PR #140, opened 2019-09-19
migrating child process exec to spawn to avoid hanging process

bfitzpat:

### DIFF
--- a/src/IntegrationUtils.ts
+++ b/src/IntegrationUtils.ts
@@ -53,6 +53,8 @@ const choiceList = [
 	dependencyIntegration
  ];
 
+ let childProcessMap : Map<string, child_process.ChildProcess>;
+
  export function startIntegration(context: vscode.Uri): Promise<boolean> {
 	return new Promise <boolean> ( async (resolve, reject) => {
 		const choice : string | undefined = await vscode.window.showQuickPick(choiceList, {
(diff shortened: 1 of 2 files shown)


src/IntegrationUtils.ts
```diff
@@ -322,41 +322,44 @@ function createNewIntegration(integrationFileUri: vscode.Uri, devMode? : boolean
...
 			});
 		}
 		if (propertyArray && propertyArray.length > 0) {
 			propertyArray.forEach(prop => {
-				commandString += ` -p ${prop}`;
+				kamelArgs.push(`-p ${prop}`);
 			});
-		}
-		console.log(`commandString = ${commandString}`);
+		}
```
lhein: whitespaces at the end

bfitzpat:
I have been using `ps ax | grep kamel` to monitor processes locally and it ends it. Will check with `ps -edf | grep kamel` though not sure why it would be different.

src/IntegrationUtils.ts
```diff
@@ -417,3 +427,28 @@ function validateDependency(text : string) {
...
+	return null;
+}
+
+// this doesn't work, but it's on the right track I think
+export function killChildProcessForIntegration(integration:string) : Promise<boolean> {
+	return new Promise( (resolve, reject) => {
+		let childProcess : any = getChildProcessForIntegration(integration);
+		if (childProcess) {
+			try {
+				childProcess.kill("SIGKILL");
```
apupier: we really need the SIGKILL? If yes, it means that there is a problem in kamel which is ignoring the SIGINT or SIGTERM.

bfitzpat: Turns out we don't need SIGKILL now. Removed it.

bfitzpat:
> working fine now :-)
> 
> By playing with it, there is another use case which I'm not sure the behavior we want:
> 
>     * start in dev mode
> 
>     * start in basic mode
>       --> currently there is still the kamel process. Should we remove it? or to ask/warn user?

If I do the same thing at the command line, the same thing happens:

* kamel run helloworld.groovy --dev
* kamel run helloworld.groovy

I'm still seeing the --dev process running there as well. I don't think this is something we're going to solve here.

bfitzpat:
Created https://github.com/apache/camel-k/issues/946 upstream to see if they have any clarity around that use case.

PR #164, opened 2019-10-16
FUSETOOLS2-10 Adding Namespace Support

bfitzpat:

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,12 @@
 						"description": "Show Status Bar messages",
 						"scope": "window"
 					},
+					"camelk.namespace": {
+						"type": "string",
+						"default": "default",
+						"description": "Default namespace to use for all Apache Camel K operations",
+						"scope": "window"
+					},
 					"camelk.tools": {
 						"type": "object",
 						"description": "CLI locations",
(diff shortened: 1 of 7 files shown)


apupier:
if I understand well it means that we don't rely anymore on the system namespace? The one used from command-line and rom kubernetes extension.

bfitzpat:
> if I understand well it means that we don't rely anymore on the system namespace? The one used from command-line and rom kubernetes extension.

It always just used the "default", but now we're being more explicit about it by providing it via the --namespace argument at the CLI

apupier:
> > if I understand well it means that we don't rely anymore on the system namespace? The one used from command-line and rom kubernetes extension.
> 
> It always just used the "default", but now we're being more explicit about it by providing it via the --namespace argument at the CLI

I don't think that it is using always "default", it is using the "current namespace" https://github.com/apache/camel-k/blame/master/README.adoc#L44
Which means that now we have the kamel command that can be unsynchronized from the Kubernetes view and from the Kamel CLI

bfitzpat:
> I don't think that it is using always "default", it is using the "current namespace" https://github.com/apache/camel-k/blame/master/README.adoc#L44
> Which means that now we have the kamel command that can be unsynchronized from the Kubernetes view and from the Kamel CLI

Well, we can make it optional then -- if it's empty it is ignored and goes with the default that is specified by the system. If it's populated, it is used and overrides it. Sound good @apupier ?

apupier:
sounds good

maybe instead of "empty" have a label like "use current namespace"?

bfitzpat:
Created https://github.com/camel-tooling/vscode-camelk/pull/186

PR #168, opened 2019-10-17
Updating readme

bfitzpat:


README.md
```diff
@@ -193,20 +193,8 @@ To access **Tooling for Apache Camel K** extension settings:
...
-
-## Starting a local Kubernetes proxy
-
-**Note:** This option is only available if you have the Minikube executable installed.
-
-1. In In the VS Code editor, open the command palette by pressing **Ctrl+Shift+P** or **F1**. 
-2. From the options, select **Apache Camel K: Start the Kubernetes proxy server**.
-
-A new Kubernetes proxy starts. It uses the Minikube executable (kubectl proxy --port=8000) and refreshes the Apache Camel K Integrations view immediately. This command creates a local proxy at the URL composed of the Proxy URL and Proxy Port set in the settings. For example, with the defaults, the proxy URL becomes **http://localhost:8000**.
+* **Namespace** - The namespace of the Kubernetes Proxy. You can select one of  two values: default and syndesis. The namespace is used to assemble the complete proxy URL and is part of the API definition in Kubernetes.
```
apupier: "of  two values" --> there is a double space

apupier: it is kubernetes namespace? or another kind of namespace? we can select only "default" or "syndesis"? why no other?
If Kubernetes namespaces, i think that by default it is installing in the current namespace, so inside Che by default it is in "che" workspace

bfitzpat: I fixed it with a follow-up PR - https://github.com/camel-tooling/vscode-camelk/pull/169/files

Issue #171, opened 2019-10-21
Fixing Cannot read property 'catch' of undefined issue


bfitzpat:
This issue has cropped up on Travis-ci

```rejected promise not handled within 1 second: TypeError: Cannot read property 'catch' of undefined

stack trace: TypeError: Cannot read property 'catch' of undefined

	at CamelKNodeProvider.<anonymous> (/home/travis/build/camel-tooling/vscode-camelk/out/CamelKNodeProvider.js:60:41)

	at Generator.next (<anonymous>)

	at /home/travis/build/camel-tooling/vscode-camelk/out/CamelKNodeProvider.js:8:71

	at new Promise (<anonymous>)

	at __awaiter (/home/travis/build/camel-tooling/vscode-camelk/out/CamelKNodeProvider.js:4:12)

	at Promise (/home/travis/build/camel-tooling/vscode-camelk/out/CamelKNodeProvider.js:56:49)

	at new Promise (<anonymous>)

	at CamelKNodeProvider.addChild (/home/travis/build/camel-tooling/vscode-camelk/out/CamelKNodeProvider.js:56:16)

	at integrationExplorer.getChildren.then (/home/travis/build/camel-tooling/vscode-camelk/out/test/suite/integrationExplorer.test.js:42:33)

rejected promise not handled within 1 second: TypeError: Cannot read property 'catch' of undefined

stack trace: TypeError: Cannot read property 'catch' of undefined

	at CamelKNodeProvider.<anonymous> (/home/travis/build/camel-tooling/vscode-camelk/out/CamelKNodeProvider.js:60:41)

	at Generator.next (<anonymous>)

	at /home/travis/build/camel-tooling/vscode-camelk/out/CamelKNodeProvider.js:8:71

	at new Promise (<anonymous>)

	at __awaiter (/home/travis/build/camel-tooling/vscode-camelk/out/CamelKNodeProvider.js:4:12)

	at Promise (/home/travis/build/camel-tooling/vscode-camelk/out/CamelKNodeProvider.js:56:49)

	at new Promise (<anonymous>)

	at CamelKNodeProvider.addChild (/home/travis/build/camel-tooling/vscode-camelk/out/CamelKNodeProvider.js:56:16)

	at integrationExplorer.getChildren.then (/home/travis/build/camel-tooling/vscode-camelk/out/test/suite/integrationExplorer.test.js:54:33)

rejected promise not handled within 1 second: TypeError: Cannot read property 'catch' of undefined

[...]

PR #172, opened 2019-10-21
issue #171 - fixing issue with CamelKNodeProvider not throwing catch

bfitzpat:


bfitzpat:
Throwing the error instead of just rejecting it seems to get around the catch problem, though I'm not sure why. Looking at https://stackoverflow.com/questions/46005954/how-to-catch-a-promise-reject it appears that perhaps returning the inner promise instead of simply trying to resolve/reject in a more disconnected state could be the answer -- trying that now.

bfitzpat:
That seems to be the better answer, relying on promise resolve/reject than an explicit throw.

Issue #143, opened 2019-09-20
Address delay between starting integration and following log


bfitzpat:
Sometimes when you start a new integration and it shows up as "running", when you go to "follow" the log, you get an error saying the container isn't quite ready:

`[ERROR] Error from server (BadRequest): container "integration" in pod "hashtagbotsecret-8f6d7f466-dl4xr" is waiting to start: ContainerCreating`

This error shows up in the Apache Camel K output channel though the integration-specific output channel appears blank. You end up with a disconnect with the error showing up in the wrong output channel. 

We either need to prevent opening the integration-specific output channel until we actually start getting log output or flip back to the main output so the user can see the error.

bfitzpat:
Fixed it with this - https://github.com/camel-tooling/vscode-camelk/pull/173
![Peek 2019-10-21 09-33](https://user-images.githubusercontent.com/530878/67220002-40ff1f80-f3e6-11e9-9ce1-8e148d9c3a19.gif)

djelinek:
verified - redhat.vscode-camelk v0.0.10

PR #173, opened 2019-10-21
issue #143 - fixing issue when starting log

bfitzpat:


bfitzpat:
Peek 2019-10-21 09-33

bfitzpat:
Essentially I watch for a particular message in the error (i.e. `waiting to start: ContainerCreating`) and simply restart every 5 seconds until we get something - even if it's just another error that stops the madness.

PR #179, opened 2019-10-23
fixed broken download of Kamel executable

lhein:

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -125,29 +125,26 @@ export async function installKamel(context: vscode.ExtensionContext): Promise<Er
 	console.log(`Downloading Apache Camel K CLI to ${installFolder}`);
 	mkdirp.sync(installFolder);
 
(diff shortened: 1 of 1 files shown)


bfitzpat:
> any ideas to write a test for it?

I really don't have any idea how to test it - do you have ideas?

apupier:
several possibilities around the subject:
- test that the latest version can be retrieved still retrieved https://github.com/camel-tooling/vscode-camelk/blob/b2fc5e94232c5ad225cbf56837e5e889163f9f75/src/installer.ts#L173
- test that the executable is available for the version retrieved and the various platform supported https://github.com/camel-tooling/vscode-camelk/blob/b2fc5e94232c5ad225cbf56837e5e889163f9f75/src/installer.ts#L128
- test that after calling https://github.com/camel-tooling/vscode-camelk/blob/b2fc5e94232c5ad225cbf56837e5e889163f9f75/src/installer.ts#L102 there is the kamel file created
- testing that it is executable, given this line, I don't understand how it can be possible for the user which is not root https://github.com/camel-tooling/vscode-camelk/blob/b2fc5e94232c5ad225cbf56837e5e889163f9f75/src/installer.ts#L138

Issue #180, opened 2019-10-23
Provide progress indication when downloading kamel/kubectl clis


bfitzpat:
On a slow connection, downloading the CLIs takes a while and the user could use some update

bfitzpat:
Working on a solution now that puts the download progress in the Apache Camel K output channel.

```
Installing kamel... 
Downloading kamel cli tool from https://github.com/apache/camel-k/releases/download/1.0.0-M3/camel-k-client-1.0.0-M3-linux-64bit.tar.gz to /home/bfitzpat/.config/Code/User/globalStorage/redhat.vscode-camelk/camelk/tools/kamel/kamel 
Downloading from: https://github.com/apache/camel-k/releases/download/1.0.0-M3/camel-k-client-1.0.0-M3-linux-64bit.tar.gz
Bytes to transfer: 23212001
Bytes transferred: 0
Bytes transferred: 16360
Bytes transferred: 16927
Bytes transferred: 33287
Bytes transferred: 34335
Bytes transferred: 50695
...
Bytes transferred: 23165087
Bytes transferred: 23165471
Bytes transferred: 23181831
Bytes transferred: 23198191
Bytes transferred: 23212001
Downloaded camel-k-client-1.0.0-M3-linux-64bit.tar.gz.
Decompressing camel-k-client-1.0.0-M3-linux-64bit.tar.gz.
done 
```

and

```
Installing kubectl... 
Downloading Kubernetes cli tool from https://storage.googleapis.com/kubernetes-release/release/v1.16.2/bin/linux/amd64/kubectl to /home/bfitzpat/.vs-kubernetes/tools/kubectl/kubectl 
Downloading from: https://storage.googleapis.com/kubernetes-release/release/v1.16.2/bin/linux/amd64/kubectl
Bytes to transfer: 46681472
Bytes transferred: 0
Bytes transferred: 466
Bytes transferred: 1855
Bytes transferred: 3244
Bytes transferred: 4633
...
Bytes transferred: 46644328
Bytes transferred: 46660712
Bytes transferred: 46677096
Bytes transferred: 46681472
Downloaded kubectl.
Decompressing kubectl.
done 
```

bfitzpat:
Created PR - https://github.com/camel-tooling/vscode-camelk/pull/201

djelinek:
verified - redhat.vscode-camelk (0.0.10)

PR #201, opened 2019-11-05
issue #180 - changing download approach and adding progress indicator

bfitzpat:

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -114,7 +114,8 @@ async function downloadAndExtract(link : string, dlFilename: string, installFold
 			let incr = progress.total > 0 ? Math.floor(progress.transferred / progress.total * 100) : 0;
 			let percent = Math.round(incr);
 			let message = `Download progress: ${progress.transferred} / ${progress.total} (${percent}%)`;
-			updateStatusBarItem(myStatusBarItem, message);
+			let tooltip = `Download progress for ${dlFilename}`;
+			updateStatusBarItem(myStatusBarItem, message, tooltip);
 		}).then(async () => {
 			extension.mainOutputChannel.appendLine(`Downloaded ${dlFilename}.`);
 			myStatusBarItem.dispose();
@@ -126,9 +127,10 @@ async function downloadAndExtract(link : string, dlFilename: string, installFold
 	return false;
 }
 
-function updateStatusBarItem(sbItem : vscode.StatusBarItem, text: string): void {
+function updateStatusBarItem(sbItem : vscode.StatusBarItem, text: string, tooltip : string): void {
 	if (text) {
 		sbItem.text = text;
+		sbItem.tooltip = tooltip;
 		sbItem.show();
 	} else {
 		sbItem.hide();
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -27,12 +27,12 @@ import * as shell from './shell';
 import * as vscode from 'vscode';
 import * as kubectlutils from './kubectlutils';
 import * as downloader from './downloader';
-
-const downloadTarball = require('download-tarball');
+import * as download from 'download';
+import decompress = require('decompress');
 
 export const kamel = 'kamel';
 export const kamel_windows = 'kamel.exe';
-export const version = '1.0.0-M2'; //need to retrieve this if possible, but have a default
+export const version = '1.0.0-M3'; //need to retrieve this if possible, but have a default
 export const PLATFORM_WINDOWS = 'windows';
 export const PLATFORM_MAC = 'mac';
 export const PLATFORM_LINUX = 'linux';
@@ -99,6 +99,39 @@ export function getPlatform() : string | undefined {
     return undefined;
 }
 
+async function downloadAndExtract(link : string, dlFilename: string, installFolder : string, extract? : boolean) : Promise<boolean> {
+	const downloadSettings = {
+        filename: `${dlFilename}`,
+        extract: true,
+	  };
+	extension.mainOutputChannel.appendLine('Downloading from: ' + link);
+	await download(link, installFolder, downloadSettings)
+		.on('response', (response) => {
+			extension.mainOutputChannel.appendLine(`Bytes to transfer: ${response.headers['content-length']}`);
+		}).on('downloadProgress', (progress) => {
+			extension.mainOutputChannel.appendLine(`Bytes transferred: ${progress.transferred}`);
+		}).then(async () => {
+			extension.mainOutputChannel.appendLine(`Downloaded ${dlFilename}.`);
+			if (extract) {
+				extension.mainOutputChannel.appendLine(`Decompressing ${dlFilename}.`);
+				await decompress(dlFilename, installFolder).then(() => {
+					extension.mainOutputChannel.appendLine('done extracting!');
+					return true;
+				}).catch((error) => {
+					console.log(error);
+					return false;
+				});
+			} else {
+				extension.mainOutputChannel.appendLine('done!');
+				return true;
+			}
+		}).catch((error) => {
+			console.log(error);
+			return false;
+		});
+	return false;
+}
+
 export async function installKamel(context: vscode.ExtensionContext): Promise<Errorable<null>> {
 	const latestversion = await getLatestCamelKVersion();
 	if (failed(latestversion)) {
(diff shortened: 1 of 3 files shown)


bfitzpat:
This puts the download progress in the Apache Camel K output channel.

```
Installing kamel... 
Downloading kamel cli tool from https://github.com/apache/camel-k/releases/download/1.0.0-M3/camel-k-client-1.0.0-M3-linux-64bit.tar.gz to /home/bfitzpat/.config/Code/User/globalStorage/redhat.vscode-camelk/camelk/tools/kamel/kamel 
Downloading from: https://github.com/apache/camel-k/releases/download/1.0.0-M3/camel-k-client-1.0.0-M3-linux-64bit.tar.gz
Bytes to transfer: 23212001
Bytes transferred: 0
Bytes transferred: 16360
Bytes transferred: 16927
Bytes transferred: 33287
Bytes transferred: 34335
Bytes transferred: 50695
...
Bytes transferred: 23165087
Bytes transferred: 23165471
Bytes transferred: 23181831
Bytes transferred: 23198191
Bytes transferred: 23212001
Downloaded camel-k-client-1.0.0-M3-linux-64bit.tar.gz.
Decompressing camel-k-client-1.0.0-M3-linux-64bit.tar.gz.
done 
```

and

```
Installing kubectl... 
Downloading Kubernetes cli tool from https://storage.googleapis.com/kubernetes-release/release/v1.16.2/bin/linux/amd64/kubectl to /home/bfitzpat/.vs-kubernetes/tools/kubectl/kubectl 
Downloading from: https://storage.googleapis.com/kubernetes-release/release/v1.16.2/bin/linux/amd64/kubectl
Bytes to transfer: 46681472
Bytes transferred: 0
Bytes transferred: 466
Bytes transferred: 1855
Bytes transferred: 3244
Bytes transferred: 4633
...
Bytes transferred: 46644328
Bytes transferred: 46660712
Bytes transferred: 46677096
Bytes transferred: 46681472
Downloaded kubectl.
Decompressing kubectl.
done 
```

bfitzpat:
Investigating moving to a progress bar as suggested by @lhein  -- https://github.com/jboss-fuse/vscode-atlasmap/blob/master/src/extension.ts#L215-L237

bfitzpat:
Moved it to the status bar with a percentage rather than putting progress into the Output channel. Tried using the progress bar but had trouble getting that approach to work for me.

lhein:
Why that? It worked fine yesterday except that the percentages where not displayed correctly due to some logical error. What made you switch to a static progress message instead?

src/installer.ts
```diff
@@ -27,12 +27,11 @@ import * as shell from './shell';
...
 import * as kubectlutils from './kubectlutils';
 import * as downloader from './downloader';
-
-const downloadTarball = require('download-tarball');
+import * as download from 'download';
 
 export const kamel = 'kamel';
 export const kamel_windows = 'kamel.exe';
-export const version = '1.0.0-M2'; //need to retrieve this if possible, but have a default
+export const version = '1.0.0-M3'; //need to retrieve this if possible, but have a default
```
apupier: would be better to have a dedicated PR for default version upgrade.

src/installer.ts
```diff
@@ -99,6 +98,43 @@ export function getPlatform() : string | undefined {
     return undefined;
 }
 
+async function downloadAndExtract(link : string, dlFilename: string, installFolder : string, extractFlag : boolean) : Promise<boolean> {
```
apupier: tabs and space mixed

src/installer.ts
```diff
@@ -126,19 +162,20 @@ export async function installKamel(context: vscode.ExtensionContext): Promise<Er
...
+	await downloadAndExtract(kamelUrl, kamelCliFile, installFolder, true)
 	.then( async (flag) => {
-		console.log(`Downloaded ${downloadFile} successfully: ${flag}`);
-		if (fs.existsSync(downloadFile)) {
-			if (shell.isUnix()) {
-				fs.chmodSync(downloadFile, '0700');
-			}
-			await config.addKamelPathToConfig(downloadFile);
-		}
+	 	console.log(`Downloaded ${downloadFile} successfully: ${flag}`);
```
apupier: space added?

src/installer.ts
```diff
@@ -233,14 +252,17 @@ export async function installKubectl(context: vscode.ExtensionContext): Promise<
...
+		console.log(`Downloaded ${downloadFile} successfully: ${flag}`);
+	}).catch ( (error) => {
+		console.log(error);
+		return { succeeded: false, error: [`Failed to download kubectl: ${error}`] };
+	});
 	await config.addKubectlPathToConfig(downloadFile);
 
 	if (shell.isUnix()) {
-		fs.chmodSync(downloadFile, '0777');
+		fs.chmodSync(downloadFile, '0700');
```
apupier: why do you need to change rights when changing the download approach?
If giving a 700, won't it means that the user needs to be an admin to be able to launch it? or that he/she launched VS Code in admin mode?

lhein: Nope....it means OWNER/USER is allowed everything, GROUP is allowed nothing, OTHERS is allowed nothing. So that change basically improves the security because it doesn't allow other users or groups to access the users home directory where this file is stored.

apupier: ah ok, seems better effectively.
remains my question:

why do you need to change rights when changing the download approach?

lhein: Why can't u ask the question you actually want to ask? "Can you make this change in a separate PR?" ;)

apupier: because if the answer is that we need this right modification to have the download progress working, it should not be in another PR. I don't think so but I'm not sure so I'm asking.

lhein:
Brian, I see another issue. The download works fine but it repeats with every startup of vscode. It seems you only check if kubectl / kamel is available on PATH variable but you do not check if those executables have been downloaded before and are residing inside that users home folder somewhere. I think that needs to be addressed. (maybe in a separate issue/pr - just let me know what you prefer)

lhein:
Brian, I created a PR on your fork which refactors this PR and switches back to using Progress Bar with correct percentage calculation and progress display. 
see https://github.com/bfitzpat/vscode-camelk/pull/4
Please check it and see if that works for you. Here its working perfectly.

bfitzpat:
> Please check it and see if that works for you. Here its working perfectly.

The reason I moved away from this code was because of the ENOENT issue, which I was unable to work around. I honestly don't see the progress bar being an improvement, but beyond that it simply doesn't work on my box. See this - https://github.com/bfitzpat/vscode-camelk/pull/4#issuecomment-550321134

bfitzpat:
> Brian, I see another issue. The download works fine but it repeats with every startup of vscode. It seems you only check if kubectl / kamel is available on PATH variable but you do not check if those executables have been downloaded before and are residing inside that users home folder somewhere. I think that needs to be addressed. (maybe in a separate issue/pr - just let me know what you prefer)

That is a separate issue, but I have seen it. It was a problem before this progress change.

src/installer.ts
bfitzpat: @lhein helped change from 0777 to 0700 a while back for the kamel CLI and I just never changed it for the kubernetes side. Since we are working in this code, it did not seem a stretch to change it at this time. I honestly don't see the need to break this out

apupier: advantages:
- smaller PR
- this modification will no tbe blocked by user comments on the progress change
- allows to have a better tracing of modifications in history
- easier to think to add an entry in changelog
- easier to understand that the modification of the permission is not needed for the progress
drawbacks:
- extra branch and PR to create

bfitzpat: @apupier it's one line in the context of another change that still affects the same code. it's staying put.

bfitzpat:
Shifted back to the progress bar after @lhein finally got it to work. Thanks Lars!

src/installer.ts
```diff
@@ -93,10 +92,48 @@ export function getPlatform() : string | undefined {
...
+						lastProg = progressamt.transferred;
+					}
+					if (progressamt.transferred >= progressamt.total) {
+						success = true;
+					}
+				});
+			resolve(success);
+		})
+		.then( () => {
+			extension.mainOutputChannel.appendLine('done!');
```
apupier: as the channel is used to log different things, i think we should specify what is done.

src/installer.ts
```diff
@@ -93,10 +92,48 @@ export function getPlatform() : string | undefined {
...
+}
+
+function downloadAndExtract(link : string, dlFilename: string, installFolder : string, extract : boolean) : Promise<boolean> {
+	return new Promise<boolean>( async (resolve, reject) => {
+		const downloadSettings = {
+			filename: `${dlFilename}`,
+			extract: extract,
+		};
+		extension.mainOutputChannel.appendLine('Downloading from: ' + link);
+		let progressTitle : string = `Download (${dlFilename})`;
```
apupier: What do you think to use "Downloading" instead of "Download"?

bfitzpat:
Though it worked locally, it does not work on travis - I reverted the change and went back to the status bar.

src/installer.ts
```diff
@@ -93,10 +92,47 @@ export function getPlatform() : string | undefined {
...
+		extract: extractFlag,
+	  };
+	extension.mainOutputChannel.appendLine('Downloading from: ' + link);
+	await download(link, installFolder, downloadSettings)
+		.on('response', (response) => {
+			extension.mainOutputChannel.appendLine(`Bytes to transfer: ${response.headers['content-length']}`);
+		}).on('downloadProgress', (progress) => {
+			let incr = progress.total > 0 ? Math.floor(progress.transferred / progress.total * 100) : 0;
+			let percent = Math.round(incr);
+			let message = `Download progress: ${progress.transferred} / ${progress.total} (${percent}%)`;
```
apupier: I know the available space is reduced in the bottom bar but indicating what is downloaded would be nice. Maybe adding it as a tooltip is a good alternative?

apupier:
> > Brian, I see another issue. The download works fine but it repeats with every startup of vscode. It seems you only check if kubectl / kamel is available on PATH variable but you do not check if those executables have been downloaded before and are residing inside that users home folder somewhere. I think that needs to be addressed. (maybe in a separate issue/pr - just let me know what you prefer)
> 
> That is a separate issue, but I have seen it. It was a problem before this progress change.

reported https://issues.jboss.org/browse/FUSETOOLS2-52 to handle this case

src/installer.ts
bfitzpat: Added tooltip

bfitzpat:
@lhein I'm going to go ahead and merge this, but if you want to provide a PR later that shifts us back to a progress bar, that would be fantastic.

PR #216, opened 2019-11-14
fixing issue discovered testing install on mac

bfitzpat:

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -233,7 +233,7 @@ function getKubectlInstallFolder(tool: string): string {
 
 export async function installKubectl(context: vscode.ExtensionContext): Promise<Errorable<null>> {
 	const tool = 'kubectl';
-	const binFile = (shell.isUnix()) ? tool : `${tool}.exe`;
+	const binFile = (shell.isUnix() || shell.isMacOS()) ? tool : `${tool}.exe`;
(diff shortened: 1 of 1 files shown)


bfitzpat:
Thanks for finding this @djelinek !

bfitzpat:
This is for https://issues.jboss.org/browse/FUSETOOLS2-59

djelinek:
I think we also need to add `|| shell.isMacOS()` here https://github.com/camel-tooling/vscode-camelk/blob/cb55607792539cf2cb759ddbf60a8bea7b7bf1cb/src/installer.ts#L266-L268

WDYT @bfitzpat ?

bfitzpat:
> And another question, is here any logic for waiting until all is downloaded? because VSCode shows this notifications even if _kubectl_ file exists

Those messages are coming from the Microsoft Kubernetes extension, not from Camel K - so I don't control them. We require the Kubernetes extension to be there so we can use the kubectl api, but we don't have much control over that side of it.

djelinek:
> Those messages are coming from the Microsoft Kubernetes extension, not from Camel K - so I don't control them. We require the Kubernetes extension to be there so we can use the kubectl api, but we don't have much control over that side of it.

OK, I see, thanks for clarification.

PR #217, opened 2019-11-15
Didact ck test

bfitzpat:
Proof of concept that does a couple of things.

1) Creates and registers a new didact tutorial to appear in the DIdact view.
2) Updated the startIntegration command so we are able to pass in values to specify either Dev Mode or Basic for a deployment type, thus enabling us to pass those values from the Didact file in a link


src/extension.ts
```diff
@@ -434,23 +434,29 @@ function handleLogViaKubectlCli(podName: string) : Promise<string> {
...
-			tutorialUri,
-			tutorialCategory);
-
+		// test to ensure didact is available 
+		const extensionId = 'redhat.vscode-didact';
+		const didactExt : any = vscode.extensions.getExtension(extensionId);
+
+		// if didact is available, register the new extension
+		if (didactExt) {
+			// command ID: vscode.didact.register
```
apupier: this comment is useless, it is completely redundant with the code

src/extension.ts
```diff
@@ -434,23 +434,29 @@ function handleLogViaKubectlCli(podName: string) : Promise<string> {
...
-			commandId,
-			tutorialName, 
-			tutorialUri,
-			tutorialCategory);
-
+		// test to ensure didact is available 
+		const extensionId = 'redhat.vscode-didact';
+		const didactExt : any = vscode.extensions.getExtension(extensionId);
+
+		// if didact is available, register the new extension
```
apupier: register the new extension? you mean register the tutorial in didact?

src/extension.ts
```diff
@@ -46,6 +47,9 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
 
 	applyUserSettings();
 
+	// call didact to register tutorial
+	await registerTutorialWithDidact(context);
```
apupier: It doesn't need to be unregistered when Camel K is deactivated?

apupier: I'm wondering if we really need to await that the tutorial is registered. I mean the Camel K extension can continue to start even if the tutorial is not registered. What do you think?

src/extension.ts
bfitzpat: Since this is an exemplar, I would rather be explicit because some people will not be familiar with the VS Code command API.

src/extension.ts
```diff
@@ -434,23 +434,29 @@ function handleLogViaKubectlCli(podName: string) : Promise<string> {
...
-
-		console.log('Tutorial URI registered: ' + tutorialUri.fsPath);
-
-		await vscode.commands.executeCommand(
-			commandId,
-			tutorialName, 
-			tutorialUri,
-			tutorialCategory);
-
+		// test to ensure didact is available
```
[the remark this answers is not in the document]
bfitzpat: I was trying to do everything in this one method to avoid complicating things for a developer simply seeking to register a new tutorial with the system.

src/extension.ts
bfitzpat: Yes, good call. Will update the comment

src/extension.ts
bfitzpat: All tutorials are unregistered and re-registered when the workbench starts, so no need to unregister. I am relying on extension activation to register any applicable tutorials for the view.

bfitzpat: I think because we are updating the workbench configuration, we need to wait for it to complete before moving on or risk collisions when other extensions also add configurations upon activation

bfitzpat:
> having a test for the tutorial would be nice, it will ensure that it stays up to date more easily. Which is one of the major maintenance pain with tutorials.

I'm not sure what we're testing here. Can you be more specific? I have added numerous tests to the didact test suite over the last week to ensure that registering tutorials works as expected. 

I still need to ensure that the Didact window remembers the state of the last tutorial that was opened, because when you close VS Code and re-open, it resets

bfitzpat:
@apupier I never heard back from you regarding the type of test you would like to see for this work. Essentially we're adding optional functionality since we aren't requiring the Didact extension be installed (we can eventually add it as a dependency like the Kubernetes extension, but for now I think it's good to leave it as optional). The changes to the actual extension functionality are minor and extensible going forward if we want to add to the functionality we're exposing in this fashion. The rest of it is simply testing to see if the Didact extension exists and if it does, is registering a Camel K tutorial. 

If there's a particular path you think we should follow, I'm all ears, but otherwise I'd like to get this in before we release the next version.

apupier:
> > having a test for the tutorial would be nice, it will ensure that it stays up to date more easily. Which is one of the major maintenance pain with tutorials.
> 
> I'm not sure what we're testing here. Can you be more specific? I have added numerous tests to the didact test suite over the last week to ensure that registering tutorials works as expected.

Have a test checking that when playing this specific Didact tutorial, things are executed correctly. The idea is to simulate a user that would follow the tutorial and ensure that the described scenario is working.

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,109 @@
...
+| [VS Code Tooling for Apache Camel K by Red Hat is installed](didact://?commandId=vscode.didact.extensionRequirementCheck&text=camelk-extension-requirement-status$$redhat.vscode-camelk&completion=Camel%20K%20extension%20pack%20available. "Checks the VS Code workspace to make sure the extension pack is installed"){.didact} | *Status: unknown*{#camelk-extension-requirement-status} 	| [Click here to install](vscode:extension/redhat.vscode-camelk "Opens the extension page and provides an install link") |
+
+## Your First Camel K Integration
+
+You can write an integration in one of several languages supported ([Groovy, Kotlin, JavaScript, Java, XML, etc.](https://camel.apache.org/camel-k/latest/languages/languages.html)), but today we're going to focus on Groovy which is one of the easiest and most popular.
```
apupier: I know the tutorial was written before it was decided to have Groovy in most focused languages.
Just wondering if we should not shift to another one in the first tutorial that we are providing.

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Step 1: Creating a Folder and Your Integration
+
+First, we need to set up our sample project. Camel K has many examples available at their [GitHub project hosting the source](https://github.com/apache/camel-k/tree/master/examples), but we're just going to use the simplest one: `simple.groovy`.
+
+You can download the file yourself from the GitHub repo and then copy it into a folder in your workspace. Or you can create a sample folder and copy in the file with the link below.
+
+- [ ] [Create a sample folder in your VS Code workspace and we will define it there.](didact://?commandId=vscode.didact.scaffoldProject&extFilePath=redhat.vscode-camelk/didact/camelk/simple-groovy-project.json&completion=Created%20simple-groovy%20project. "Creates a folder and copies in simple.groovy"){.didact}
```
apupier: this action did nothing for me when I add several folders/files in the workspace

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Step 3: Deploying the Integration
+
+![Camel K Start Integration menu](https://raw.githubusercontent.com/camel-tooling/vscode-camelk/master/images/camelk-start-integration-popup-menu.jpg){.imageRight}
+
+The `Tooling for Camel K` extension offers several tools to get your new integration started. 
+
+First, you can right-click the `simple.groovy` file and select `Start Apache Camel K Integration`. That will provide a drop-down in the Command palette area with a number of deployment options. In this case, select the `Dev Mode - Apache Camel K Integration in Dev Mode` option. 
+
+- [ ] [Start the simple.groovy integration in Dev Mode](didact://?commandId=camelk.startintegration&projectFilePath=first-camelk-integration/src/simple.groovy&text=Dev%20Mode&completion=Deployed%20the%20simple.groovy%20file%20in%20Dev%20mode. "Deploys the simple.groovy file in 'Dev mode'"){.didact}
```
apupier: this is not working if the file wasn't created in first-camelk-integration folder. it is failing silently in this case.
Would be nice to provide at least a warning or explain in text that it needs to follow the pattern (the best would be to detect the simple.groovy file in the workspace but looks alike a new feature and can be done in another iteration)

src/extension.ts
apupier: I don't understand why it is an exemplar.

Even if exemplar, providing comments that are duplicating exactly the code is giving no information to users. It is creating duplicating information that can go unsynchronized.

src/extension.ts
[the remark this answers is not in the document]
apupier: I think that it is simpler to have explicit naming with method names than several comments to explain what is the purpose of the code.

to quote Ron Jeffries author of Extreme Programming Installed and Extreme Programming Adventures in C#

Expressiveness to me includes meaningful names, and I am likely to change the names of things several times before I settle in. With modern coding tools such as Eclipse, renaming is quite inexpensive, so it doesn’t trouble me to change. Expressiveness goes beyond names, however. I also look at whether an object or method is doing more than one thing. If it’s an object, it probably needs to be broken into two or more objects. If it’s a method, I **will always use the Extract Method refactoring** on it, **resulting in one method that says more clearly what it does, and some submethods saying how it is done.**

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,109 @@
...
+
+While our integration is running in `Dev Mode`, we can modify it and see those changes reflected in the deployed integration. Let's try that now.
+
+Change the message sent to the `.simple()` command of the Camel route in quotes to `We just changed our first Camel K integration while it was running!`. Save the file and see what happens in the Output channel.
+
+## Step 5: Managing our Integration
+
+We can see what integrations we currently have running in our Minikube system in the `Apache Camel K Integrations` view in the Explorer activity (Ctrl+Shift+E).
+
+- [ ] Open the `Apache Camel K Integrations` view
```
apupier: improvement idea for another iteration: have a anction to make the Apache Camel K Integrations view visible

didact/camelk/first-integration.md
paoloantinori: I don't think we stricly need to change what we have. We can add more eventually, but I don't see reasons to change the current one.

didact/camelk/first-integration.md
paoloantinori: +1

VSCode has showed already obscure implicit assumptions related to folders a project roots. 

Ideally we want to push warnings up to Didact itself or reason on macros to be used by the document author to help in recurring situations like these.

src/extension.ts
paoloantinori: > this comment is useless, it is completely redundant with the code

https://github.com/thoughtbot/guides/tree/master/code-review

paoloantinori: > Even if exemplar, providing comments that are duplicating exactly the code is giving no information to users. It is creating duplicating information that can go unsynchronized.

this statement is true for every comment written anywhere, and it's not an strong enough argument for not using comments.

"Clear enough code shouldn't require comments" is a very extreme position that developer can have but that shouldn't try to force on others.

apupier: > "Clear enough code shouldn't require comments" is a very extreme position that developer can have but that shouldn't try to force on others.

in this case, this is not even this statement. it is for duplicating information. if the comment is providing more context, more information, why not. The pointed one is an exact duplication, there is no additional information provided.

apupier: > > this comment is useless, it is completely redundant with the code
> 
> https://github.com/thoughtbot/guides/tree/master/code-review

sorry should formulate:

this comment is completely redundant with the code and doesn't provide any other piece of information. Given that it can be safely removed: no information will be lost and the code will be clearer.

src/extension.ts
apupier: it is possible to disable an extension, i think that in this case, the tutorial will still be registered even though it won't be available anymore.

apupier: ok fine for a first iteration.

this is something that will need to be improved in the future because it is breaking the whole asynchronuous effort that has been done in VS Code to provide a more responsive UI.

In case one extension is providing a huge amount of tutorial, I fear we will notice the slowness.

didact/camelk/first-integration.md
apupier: a reason would be to provide a tutorial on one of the language that Fuse is focusing/planned to support first (Java, xml, Javascript).

i'm fine to keep Groovy, it is just that we are not in phase of Fuse priorities.

bfitzpat:
> there are several links to camel.apache.org website, for which it is asked if we really want to open them and propose to configure the Trusted Domains. Wondering if we can define trusted domains to avoid user to do it. (seems a candidate more for another iteration

I'm not sure we can (or should) automate this. But it's beyond the scope of this particular PR for sure. The only real documentation I've seen is in the September update: https://code.visualstudio.com/updates/v1_39#_improved-trusted-domains-setting

didact/camelk/first-integration.md
bfitzpat: Good idea for another PR. There is no generic way to do it I have found, so this sort of functionality must be implemented on a per-extension basis - https://github.com/redhat-developer/vscode-didact/issues/4

didact/camelk/first-integration.md
bfitzpat: I don't believe that is the case, though it may have been difficult to tell at the time. I suspect that the folder/file structure was created beneath the first folder in the workspace whether you noticed it or not.

didact/camelk/first-integration.md
bfitzpat: Easy enough to change languages, but I was going with what was suggested by the runtime we're supporting and simply migrating the suggested Camel K runtime example from the documentation https://camel.apache.org/camel-k/latest/running/running.html 

In the long run, I anticipate that these files will be written/modified by our documentation team, which means we will have less direct input on their content.

bfitzpat:
> > > having a test for the tutorial would be nice, it will ensure that it stays up to date more easily. Which is one of the major maintenance pain with tutorials.
> > 
> > 
> > I'm not sure what we're testing here. Can you be more specific? I have added numerous tests to the didact test suite over the last week to ensure that registering tutorials works as expected.
> 
> Have a test checking that when playing this specific Didact tutorial, things are executed correctly. The idea is to simulate a user that would follow the tutorial and ensure that the described scenario is working.

So an additional test that walks through the various commands that are called/used by the tutorial itself to ensure that they are working properly? Sure. I guess I can see that.

apupier:
> > > > having a test for the tutorial would be nice, it will ensure that it stays up to date more easily. Which is one of the major maintenance pain with tutorials.
> > > 
> > > 
> > > I'm not sure what we're testing here. Can you be more specific? I have added numerous tests to the didact test suite over the last week to ensure that registering tutorials works as expected.
> > 
> > 
> > Have a test checking that when playing this specific Didact tutorial, things are executed correctly. The idea is to simulate a user that would follow the tutorial and ensure that the described scenario is working.
> 
> So an additional test that walks through the various commands that are called/used by the tutorial itself to ensure that they are working properly? Sure. I guess I can see that.

yes

bfitzpat:
Long story short, this will not get merged for this release and I will hold off until next year when the Didact work is a bit more fully baked.

bfitzpat:
I will create a new "exemplar" on how to register a new didact tutorial and include it as a sub-project of the didact project rather than attempting to use the Camel K extension as the first go.

bfitzpat:
Closed in favor of https://github.com/camel-tooling/vscode-camelk/pull/385

Still some work to do there, but good progress and tried to pull in some of the feedback from this PR

PR #223, opened 2019-11-21
FUSETOOLS2-60 fixing treeview images so they appear again

bfitzpat:

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -32,7 +32,7 @@ import * as config from './config';
 export let mainOutputChannel: vscode.OutputChannel;
 export let myStatusBarItem: vscode.StatusBarItem;
 
-let camelKIntegrationsProvider = new CamelKNodeProvider();
+let camelKIntegrationsProvider : CamelKNodeProvider;
 let outputChannelMap : Map<string, vscode.OutputChannel>;
 let showStatusBar : boolean;
 let camelKIntegrationsTreeView : vscode.TreeView<TreeNode>;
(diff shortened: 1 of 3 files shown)


bfitzpat:
> What about adding a test checking that getIconForPodStatus is returning an existing file?

Good idea so we don't hit this problem again. :) Added a test and restructured the method slightly for easier testing purposes.

src/extension.ts
```diff
@@ -32,18 +32,24 @@ import * as config from './config';
...
 
-let camelKIntegrationsProvider = new CamelKNodeProvider();
+let camelKIntegrationsProvider : CamelKNodeProvider;
 let outputChannelMap : Map<string, vscode.OutputChannel>;
 let showStatusBar : boolean;
 let camelKIntegrationsTreeView : vscode.TreeView<TreeNode>;
 let eventEmitter = new events.EventEmitter();
 const restartKubectlWatchEvent = 'restartKubectlWatch';
 
+let stashedContext : vscode.ExtensionContext;
```
lhein: Just a suggestion:
Why not simply call it context and the getter then getContext() instead of mixing stashedContext and sharedContext? Not a prio thing but would make the code look cleaner.

apupier: agree, `stashedContext` <-> `shareContext`, would be best to use always the same.
Would be also better to use directly `context` but we have this naming used as a parameter of the method, as we are not in a Class we cannot use `this.context` to reference the attribute. (maybe there is a tweak with TypeScript language but I have not found it)

PR #231, opened 2019-12-03
FUSETOOLS2-52 - retrieving client instead of runtime version of kubectl

bfitzpat:

### DIFF
--- a/src/downloader.ts
+++ b/src/downloader.ts
@@ -4,29 +4,11 @@ import * as tmp from 'tmp';
 
 import { succeeded, Errorable } from './errorable';
 
-type Dictionary<T> = {
-    [key: string]: T
-};
-
-module Dictionary {
-    export function of<T>(): Dictionary<T> {
-        return {};
-    }
-}
-
 type DownloadFunc =
     (url: string, destination?: string, options?: any)
          => Promise<Buffer> & stream.Duplex; // Stream has additional events - see https://www.npmjs.com/package/download
 
 let download: DownloadFunc | undefined;
-
-const DOWNLOAD_ONCE_STATUS = Dictionary.of<DownloadOperationStatus>();
-
-enum DownloadOperationStatus {
-    Queued = 1,
-    Completed = 2,
-    Failed = 3,
-}
 
 function ensureDownloadFunc() {
     if (!download) {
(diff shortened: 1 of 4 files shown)


bfitzpat:
Essentially the change here is to retrieve the CLI version (--client) instead of the client and runtime version, which is the default. This gets us what we are looking for as far as making sure that the CLI is installed properly and prevents redownloading the kubernetes CLI unnecessarily.

lhein:
I agree, but that is not the scope of this PR.

apupier:
> I agree, but that is not the scope of this PR.

It is a non-regression test for the issue that is fixed by this PR. It seems to fit well in the scope of this PR.
Or maybe I misunderstand the reason of the code modification?

lhein:
Or maybe I misunderstood the reason of this PR. I thought its just a change for getting the correct version number.

bfitzpat:
As to the purpose of this change, we have been downloading the kubectl cli every single time, regardless of whether it's already there. I was using the method in question to validate that the CLI was already there, but it was not returning a known result in the case where the Kubernetes system wasn't already running, even if the CLI was installed. 

The method now returns the version of the CLI explicitly, getting around the case where the runtime isn't actually running but the CLI was downloaded properly. So now it only downloads if it doesn't find the CLI at all, which is what we wanted in the first place. This eliminates repeatedly downloading kubectl over and over again every time we start the IDE.

bfitzpat:
That said, I have no clue how to ensure that the installer doesn't automatically redownload each time. The tests run in an environment where the install happens once only and we're only testing to ensure that the CLIs are available after the extension is activated. 

@apupier I'm open to suggestions on how to test this but don't have any ideas at this time.

apupier:
I imagine:
- spy the method that is responsible for the effective download
- call twice the method that is "checking if available then download", maybe it is the method that is "accessing" one of the kubectl call.
- check that the spied method has been called a single time

bfitzpat:
>     * spy the method that is responsible for the effective download
> 

Any examples of how to spy a method effectively?

apupier:
> > ```
> > * spy the method that is responsible for the effective download
> > ```
> 
> 
> Any examples of how to spy a method effectively?

https://www.npmjs.com/package/ts-mockito#call-count-verification
https://www.npmjs.com/package/ts-mockito#spying-on-real-objects

bfitzpat:
Ok @apupier , I don't believe I can mock anything in this case. We're not creating classes, we're calling exported functions from modules. Essentially I'd need to somehow mock "installKubectl" from the installer module. Mocking seems to work great for objects, but not for modules.

bfitzpat:
Cleaned up downloader.ts while we were working in the installer/download code to eliminate the issue mentioned in FUSETOOLS2-58 with Sonar

But I don't think we can mock a module, so we may be stuck on that front. Still investigating

bfitzpat:
Added test that uses a sinon spy (https://sinonjs.org/releases/v7.5.0/spies) to test a module/function to see how many times it gets called.

lhein:
Looks good to me, Brian.

bfitzpat:
Made one last change since we are resetting the spy count, the number should be zero. It was already installed by this point, which means it should not install a second time if we activate the extension a second time.

PR #232, opened 2019-12-03
FUSETOOLS2-58 - removing while true for sonar

bfitzpat:


bfitzpat:
Closing as not the right approach

PR #244, opened 2019-12-20
FUSETOOLS2-81 - support pure Java language with standalone Camel K Java

apupier:
files

- the Camel K files requires to contain "camel" inside it (same
constraint to have Camel URI support)
- currently providing dependencies only from camel-core-engine. Camel K
can pull some other dependencies but needs some work on kamel inspect to
leverage that
- not smart dependencies declaration all artifacts are hardcoded which
will make it hard to maintain on the long-run but should be fine for a
first iteration

/!\ it requires vscode-java 0.55.0 https://download.jboss.org/jbosstools/jdt.ls/staging/?C=M;O=D (planned to be released on the 23rd December)

will need to think mor eon it but I think that the best to test it is to write a UI test checking for completion. I will surely delegate to QE in this case.

Screenshot from 2019-12-20 14-10-56

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the "vscode-camelk" extension will be documented in this 
 
 ## 0.0.11
 
-- TBD
+- Pure Java language support with standalone Camel K Java files (requires vscode-java 0.55.0 and that the file contains the word `camel`)
 
 ## 0.0.10
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -915,7 +915,6 @@
             "version": "1.1.3",
             "resolved": "https://registry.npmjs.org/define-properties/-/define-properties-1.1.3.tgz",
             "integrity": "sha512-3MqfYKj2lLzdMSf8ZIZE/V+Zuy+BgD6f164e8K2w7dgnpKArBDerGYpM46IYYcjnkdPNMjPk9A6VFB8+3SKlXQ==",
-            "dev": true,
             "requires": {
                 "object-keys": "^1.0.12"
             }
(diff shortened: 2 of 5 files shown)


bfitzpat:
So this change is to support a standalone Java file without a maven project around it?

apupier:
> So this change is to support a standalone Java file without a maven project around it?

yes

bfitzpat:
What are the requirements that need to be met for it to work? Any additional configuration like you showed before with adding the package json to the preferences somewhere?

apupier:
> What are the requirements that need to be met for it to work? Any additional configuration like you showed before with adding the package json to the preferences somewhere?

- the Camel K ava files requires to contain "camel" inside it (same
constraint to have Camel URI support)
- the Camel K java files needs to not be in a Maven/Gradle project (but in this case, the completion will come from Maven/Gradle)
- currently with vscode-java 0.55.0, multi-root workspace folder are not supported for this feature

bfitzpat:
>     * currently with vscode-java 0.55.0, multi-root workspace folder are not supported for this feature

Do we know if multi-root workspaces are going to be supported soon? All of VS Code is going multi-root pretty much, so it seems it would have to

apupier:
> Do we know if multi-root workspaces are going to be supported soon? All of VS Code is going multi-root pretty much, so it seems it would have to

I think I never used multi-root workspaces myself yet.
As far as I know it is planned to be worked quite soon as it was mentioned to be worked on [after the referencedLibraries](https://github.com/eclipse/eclipse.jdt.ls/issues/1303#issue-537404705) features (that I'm using in this PR which was merged yesterday)

apupier:
version 0.55.1 is available online https://marketplace.visualstudio.com/items?itemName=redhat.java

PR #257, opened 2020-01-06
FUSETOOLS2-187 - list only top level dependency for Camel K Java

apupier:
- use maven dependency:copy-dependencies goal
- use a specific pom.xml with the top-level dependency defined
- keep the version as a parameter, I think it will be eaiser to spot
where to change it and maybe create a parameter to let user modify it in
the future
- the drawback is that it now requires Maven to be installed to benefit
from java Language completion, it doesn't affect other functionalities

sharing this PR to start thinking if it is fine following this way, as it is raising these questions:
Are we ready to have Maven installed ont the system as a prerequisite to benefit from Java standalone?
Or we try to provide a Maven wrapper?
Or we keep the burden to maintain a long list of artefacts?
Or we download at compile time the artefacts? But which means that it will become complicated to be dynamic in the future and have a bigger deliverables?


README.md
```diff
@@ -16,7 +16,8 @@ For more information about Camel K, be sure to check out its [documentation](htt
 In order to use our **Tooling for Apache Camel K** extension for VS Code, you must have the following software in place:
 
 * An instance of Apache Camel K running on a Kubernetes or an OpenShift cluster that is  accessible from your system on your network. You must also have Minikube (or the Kubernetes CLI) installed. See the Apache Camel K installation page for details: (https://camel.apache.org/camel-k/latest/installation/installation.html). 
-* Microsoft VS Code installed. You can get the most recent version from (https://code.visualstudio.com/) for your chosen operating system. 
+* Microsoft VS Code installed. You can get the most recent version from (https://code.visualstudio.com/) for your chosen operating system.
+* To benefit from Java Language completion on standlaone files, Maven needs to be available on command-line
```
bfitzpat: Misspelled "standlaone" - should be "standalone"

bfitzpat:
Obviously installing Maven would be the easiest answer, but I'm not sure it's the best. It drags a long list of expectations with it. It may be very useful for Java-based projects, but we should look to see if the runtimes we are trying to support are using maven. In the Camel K case, I'm not sure I've seen an entire project -- just files. And they like the free-floating file approach it seems.

Until we have a better idea of what the Camel K runtime folks are proposing as far as project organizational best practices, it might be best to put this on the back burner or leave it as optional?

apupier:
I think the Camel K project won't use Maven. Even if they are using Maven, it means that we won't need anymore this specific development as we will be able to rely on default Maven support.
Therefore I think that Camel K users does not need Maven for something else when relying on this feature.

> Until we have a better idea of what the Camel K runtime folks are proposing as far as project organizational best practices, it might be best to put this on the back burner

they are pushing back usage of Maven fo r their project for a long time. I don't think they are ready to change that so I would not await a move on this side. I think we need to find a solution on our side. (which can be to let as-is and be ready to maintain the full list of dependencies manually)

> or leave it as optional?

Can you precise what you mean?
In the current state, if Maven is not available, all Camel K features will work except the Java Language support.

bfitzpat:
Let's table this and discuss it at one of our upcoming team meetings. I think the way you've already described things, it's "optional" in the sense that it's not needed UNLESS you want the Java Language Support to work correctly. And that's an important caveat.

Ultimately, though I understand the runtime team's aversion to full Maven support, I think we'll need to support it on our side (as you point out) for Java language support. We should find a way to enable that easily for the user, perhaps as a menu action to "enable" it for a particular folder/project since VS Code doesn't really have the concept of "projects" -- only "folders" - so it becomes context dependent.

lhein:
I agree. Lets discuss that in a call.

resources/maven-project/.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright 2007-present the original author or authors.
```
apupier: copyright is strange but it is what is generated by Takari wrapper

apupier:
it was a lot easier than what I thought :-)
no more need of Maven installation locally.

resources/maven-project/.mvn/wrapper/maven-wrapper.properties
```diff
@@ -0,0 +1,2 @@
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.6.3/apache-maven-3.6.3-bin.zip
```
bfitzpat: So we will have to keep these urls in sync each time maven releases new versions upstream?

apupier: this is a fixed URL. When there is a new version of Maven, the old ones stays at their places. So when there is a new version of Maven, we can live with this one as long as we want.

This is generated by Takari Wrapper. We should not maintain this url directly ourself. When there is a new version of Maven, usually, there is a new version of takari and we can call
`mvn -N io.takari:maven:X.X.X:wrapper` in `resources/maven-project` that will take care of updating all of what is needed.
see https://github.com/takari/maven-wrapper#maven-wrapper

apupier:
> Are there changes required for the tests to ensure that the maven wrapper is in place?

when we will have basic UI tests checking for completion, it will use the Maven wrapper by default as I don' tthink we install Maven on the CI for VS Code Camel K.
But for now there is nothing, agree it sounds more and more criticla to have these tests. Rising the point, makes me rethinking about it. Thanks to this Maven wrapper, it should be easier to setup the test, so I can have a look to do it in this PR.

resources/maven-project/.mvn/wrapper/maven-wrapper.properties
bfitzpat: Cool. So long as they are pretty permanent and have some well-defined ways of updating them down the line, life should be good. Thanks

apupier:
so test passign locally from command-line and from VS Code UI. With mvn available on command-lien and without mvn availabel on command-line.
but failing on Ci.
Based on timeout failing, seems that the Java server is started but the completion is nto available 10 seconds after the Java server has started

apupier:
it was a matter of timeout.
The test is taking 36 seconds on a completely blank workspace (no Maven, no dependencies of basic maven plugins): https://travis-ci.org/camel-tooling/vscode-camelk/builds/639889633#L399

we now have a working test!

PR #263, opened 2020-01-13
FUSETOOLS2-197 - escape spaces in path

apupier:
as java extension requires it.
Java extension is using glob but cannot found a library that is covering
all possible escaped characters so doing it for the only one that is
spotted for now. This use case is the default on Mac OS.

### DIFF
--- a/src/JavaDependenciesManager.ts
+++ b/src/JavaDependenciesManager.ts
@@ -43,7 +43,7 @@ export function downloadJavaDependencies(extensionStorage:string): string {
 }
 
 export function updateReferenceLibraries(editor: vscode.TextEditor | undefined, destination:string) {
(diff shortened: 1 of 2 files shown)


apupier:
@djelinek Can you validate that it is working on Mac please?

djelinek:
@apupier Hi, I re-tested reported bug again and it is working as expected with blank space inside path.. I am very sorry for my mistake.. It had to been caused by some weirdness inside my local vscode settings and history of installed extensions don't know for sure. I had to cleaup all vscode stuff and try again because with this PR modifications it was not working too but if I switched back to latest extension snapshot it was working fine now..

apupier:
ok so closing this PR

PR #273, opened 2020-01-22
FUSETOOLS2-205 - avoid infinite loop when no Kubernetes instance

apupier:
available

- start/stop kubectl watch depending if the Apache Camel K view is
visible or not

TODO:
- validate that this watch is used only for Apache Camel K view --> Edit: validated that it is
- write a test --> Edit: test provided

For Future:
- there might a nicer way to handle the running kubectl watch process:
have delay between relaunch in case of timeout, avoid mix of event to
start and direct access to stop

note there is still an infinite loop when teh Apache Camel K view is opened (but at least there is a workaround now and it won't be the default)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -29,6 +29,7 @@ import * as kamel from './kamel';
 import * as kubectlutils from './kubectlutils';
 import * as config from './config';
 import { downloadJavaDependencies, updateReferenceLibraries } from './JavaDependenciesManager';
+import { ChildProcess } from 'child_process';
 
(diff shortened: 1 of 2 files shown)


apupier:
trying to write a test but the spy is always reporting no call!
https://github.com/apupier/vscode-camelk/tree/FUSETOOLS2-205-avoidLoop-testattempt

:-'(

bfitzpat:
At least it works when the view is closed. That's a definite improvement.

Issue #270, opened 2020-01-21
When activating extension, verify version on install


bfitzpat:
Discovered a bug -- though we may check to see if we get back a version, we're not actually validating that the version is the most recent. 

This needs to change 
https://github.com/camel-tooling/vscode-camelk/blob/0658af265a1c1c54b25121cb443afdcbffbb02c0/src/extension.ts#L367

We are not actually checking the version returned -- just that there IS a version returned.

bfitzpat:
The workaround is to delete the `kamel` executable in /.config/Code/User/globalStorage/redhat.vscode-camelk/camelk/tools/kamel

djelinek:
verified - redhat.vscode-camelk 0.0.14-413 _(snapshot)_, VS Code 1.43.2

PR #274, opened 2020-01-22
FUSETOOLS2-206 if existing version is found and not up to date, ensur…

bfitzpat:
…e update

### DIFF
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -137,6 +137,34 @@ function updateStatusBarItem(sbItem : vscode.StatusBarItem, text: string, toolti
 	}
 }
 
+export async function checkKamelNeedsUpdate(versionToUse? : string): Promise<boolean> {
+	return new Promise<boolean>( async (resolve) => {	
+		if (!versionToUse){ 
+			const latestversion = await getLatestCamelKVersion();
+			if (failed(latestversion)) {
+				console.error(latestversion.error);
+				resolve(true);
+				return true;
+			}
+			versionToUse = latestversion.result.trim();
+		}
+		await checkKamelCLIVersion().then((currentVersion) => {
+			if (versionToUse && versionToUse.toLowerCase() === currentVersion.toLowerCase()) {
+				// no need to install, it's already here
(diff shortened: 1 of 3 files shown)


bfitzpat:
https://github.com/camel-tooling/vscode-camelk/issues/270
https://issues.redhat.com/browse/FUSETOOLS2-206

bfitzpat:
works in all three cases -- if the kamel cli is missing, it is downloaded; if it is a different version, it is downloaded to match the current version (which wasn't the case before); and if the versions match, it is not downloaded

src/installer.ts
```diff
@@ -145,10 +173,10 @@ export async function installKamel(context: vscode.ExtensionContext): Promise<Er
 
 	let versionToUse = latestversion.result.trim();
 
-	await checkKamelCLIVersion().then((currentVersion) => {
-		if (versionToUse.toLowerCase() === currentVersion.toLowerCase()) {
+	await checkKamelNeedsUpdate(versionToUse).then((needsUpdate) => {
+		if (needsUpdate) {
 			// no need to install, it's already here
```
apupier: this comment is outdated. In this case, the version needs to be downloaded and updated

apupier:
with this PR i tmeans that now a user cannot provide a specific version of the kamel executable? (to support a specific version of deployment for instance)

bfitzpat:
> with this PR i tmeans that now a user cannot provide a specific version of the kamel executable? (to support a specific version of deployment for instance)

Correct. We will always default to the most recent version supported by the tooling with this change. If you have a suggestion on how to offer an additional state, I'm open to suggestions. The only thing I can think is to offer another preference that essentially says "turn off automatic update of CLI" and works around the version check.

bfitzpat:
> given that it is to fix a bug, it would be nice to add a test at least to cover the case of the detected bug

Test added.

bfitzpat:
Created a new issue to handle adding a way for the user to avoid the auto-upgrade - https://github.com/camel-tooling/vscode-camelk/issues/280

PR #302, opened 2020-02-11
FUSETOOLS2-215 - Provide Task to launch Camel K Integration

apupier:
- Task metamodel for most of the kamel attributes possible
- default task to launch indev mode
- using a ShellExecution to allow user to check the command used and the
output of it

improvements possibilities which I think can be handled in another iteration:
- more precise model definition for attributes having a <key>=<value>
list such as properties, environment variables (but need to find how to
specify that in the package.json...)
- improve completion for values of each attributes. Either using a
specific VS Code completion provider, or Language Server for tasks.json
files or defining more in package.json if it is possible (but it won't
be dynamic in that case)
- auto-generate metamodel based on kamel definition (but seems a really
hard task)
- use full kamel path if kamel not available on bin path
- Provide option in "Start Apache Camel K integration" to use a task
when users want a multi-attributes launch
- provide a Camel K task template available as completion in tasks.json

multiattributesCamelKLaunch

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,52 @@ To update the state of your currently deployed integrations, hover over the **Ap
 
 ![Apache Camel K Integrations view - Refresh](images/camelk-integrations-view-refresh-action.jpg)
 
+## Starting a new Camel K Integration with multiple parameters
+
+Using the right-click menu is the easiest way to start a new integration and works well for simple cases. When the Camel K integration requires more configuration, you can set that up using a Task.
+
+To create a new Task, you have a few options.
+
+1. From command palette, call the `Tasks: Configure Task` command. This will open the tasks.json file where you can enter new task info by hand (with some auto-completion support in the editor). If you already have a tasks.json but no Camel K tasks, you can create a new task with type `camel-k` in the tasks.json which is automatically opened.
+2. If you don't have any Tasks configured yet, you can use the `camel-k: Start in dev mode Camel K integration opened in active editor` command, which will template an initial Camel K task for you.
+
+You will end up with something like the following with a new, empty Camel K task specified:
+
+```
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558 
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "type": "camel-k",
+            "dev": true,
+            "file": "${file}",
+            "problemMatcher": []
+        }
+    ]
+}
+```
+
+Note: We recommend adding a `label` attribute to more easily find your task in the task list when you try to run it.
+
+Once you've created your Camel K task, you can use auto-complete to explore the various parameters available. The following example launches the integration in `--dev` mode and provides a couple of dependencies:
+
+```
+{
+    "type": "camel-k",
+    "label": "Run RestWithUndertow and Dependencies",
+    "dev": true,
+    "file": "./examples/RestWithUndertow.java",
+    "problemMatcher": [],
+    "dependencies": [ "camel-rest", "camel-undertow" ]
+}
+```
+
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -102,6 +102,21 @@ To update the state of your currently deployed integrations, hover over the **Ap
 
 ![Apache Camel K Integrations view - Refresh](images/camelk-integrations-view-refresh-action.jpg)
 
+## Starting a new Camel K Integration with multiple parameters
+
+Using the right-lick is very convenient to get started and in simple cases. When the Camel K integration requires more configuration, this can be provided using a Task.
```
bfitzpat: right-click? (right-lick is a fun typo though)

bfitzpat:
This was the older file I tried and failed to deploy with the first task (and at command line) - https://gist.github.com/bfitzpat/4156836be6b028115054581998d4fd7e

bfitzpat:
Here are my recommended changes to the readme:

```
## Starting a new Camel K Integration with multiple parameters

Using the right-click menu is the easiest way to start a new integration and works well for simple cases. When the Camel K integration requires more configuration, you can set that up using a Task.

To create a new Task, you have a few options.

1. From command palette, call the `Tasks: Configure Task` command. This will open the tasks.json file where you can enter new task info by hand (with some auto-completion support in the editor). If you already have a tasks.json but no Camel K tasks, you can create a new task with type `camel-k` in the tasks.json which is automatically opened.
2. If you don't have any Tasks configured yet, you can use the `camel-k: Start in dev mode Camel K integration opened in active editor` command, which will template an initial Camel K task for you.

You will end up with something like the following with a new, empty Camel K task specified:

```
{
    // See https://go.microsoft.com/fwlink/?LinkId=733558 
    // for the documentation about the tasks.json format
    "version": "2.0.0",
    "tasks": [
        {
            "type": "camel-k",
            "dev": true,
            "file": "${file}",
            "problemMatcher": []
        }
    ]
}
```

Note: We recommend adding a `label` attribute to more easily find your task in the task list when you try to run it.

Once you've created your Camel K task, you can use auto-complete to explore the various parameters available. The following example launches the integration in `--dev` mode and provides a couple of dependencies:

```
{
    "type": "camel-k",
    "label": "Run RestWithUndertow and Dependencies",
    "dev": true,
    "file": "./examples/RestWithUndertow.java",
    "problemMatcher": [],
    "dependencies": [ "camel-rest", "camel-undertow" ]
}
```

When you're done, save the file and call command from palette `Tasks: Run Task`. You will see the command listed with the label that you provided previously.

After running the camel-k `Task`, a terminal will open where you can see the command used and any execution result.
```

apupier:
> Running it again does not start it again, even after removing the failed integration. I had to stop the task terminal window and then I was able to start it again.

In fact, when using --dev, the command is not returning.

Consequently the "task is still considered as active, I have a notification taking care of that (which is provided by default VS Code workflow):
Screenshot from 2020-02-12 16-13-10

bfitzpat:
> In fact, when using --dev, the command is not returning.

That's a good point. Do we want to default to --dev mode being set to true right off the bat? Or do you think the information pop up is enough? I missed the pop up if it was there, which is why I asked.

apupier:
VS Code is providing options to use dedicated terminal via the "presentation" attribute

> > In fact, when using --dev, the command is not returning.
> 
> That's a good point. Do we want to default to --dev mode being set to true right off the bat? Or do you think the information pop up is enough? I missed the pop up if it was there, which is why I asked.

I tend to think that using --dev by default is better as it is one of the killer feature.
I think the inofrmation pop-up is enough for regular users of VS Code as it is in the classic flow. For others, it is effectively not the best.
I tried to use the presentation->panel-> new option and isBackground->true but there is still the same behavior

Issue #39, opened 2019-07-09
Enhance the console output for followed Camel-k logs


bfitzpat:
> Per apupier - "maybe in a next iteration but it seems that Kubernetes extension is providing a nice Log viewer as a view or an editor. it might be better than using the podOutputChannel (and still cannot test but it seems that in case of several pod, the log will be mixed in it, isn't it?)"
> 
> Me - So far as I  know, we can't rely on elements from other extensions -- has that changed?
> 
> Aurelien - you're right. Then maybe in a next iteration, duplicate the log editor code (or find a way to extract it as a normal javascript extension and submit it)
 
Good thoughts here - we will have to see what we can do to improve it.

bfitzpat:
Looks like this is one of the issues - https://github.com/Azure/vscode-kubernetes-tools/issues/460

bfitzpat:
And perhaps this is the webview they've added - https://github.com/Azure/vscode-kubernetes-tools/blob/master/src/components/logs/logsWebview.ts

bfitzpat:
We cannot add color to the output channel apparently. (Tried using the Chalk https://www.npmjs.com/package/chalk library and it's just ignored.)

apupier:
> We cannot add color to the output channel apparently. (Tried using the Chalk https://www.npmjs.com/package/chalk library and it's just ignored.)

seems to be related to this feature request https://github.com/Microsoft/vscode/issues/571

bfitzpat:
Though we could go with an approach similar to what the Kubernetes extension is doing with webview, I would rather not open a separate window (essentially an editor) when we already have the output channels down below. It may be the only way to get "prettier" text for the logs, but it seems like a workaround more than anything else... 

And I don't believe the Colorizer approach (https://code.visualstudio.com/api/extension-capabilities/theming) will work, since it applies to code syntax coloring in editors and views, not specifically the output channels.

bfitzpat:
Looks like spinning out the logs into separate windows like the Kubernetes extension may actually give us the functionality we are looking for. A new request for having more information more readily available (perhaps something like https://twitter.com/LostInTangent/status/1207718123241033728 or https://twitter.com/LostInTangent/status/1211342492324388864).

If we spin up these log windows as separate from the output view, we can move them around like we want. We can also look at finding a way to get easier access to the output log. 

![image(3)](https://user-images.githubusercontent.com/530878/74189232-45597d80-4c0d-11ea-85f2-c2edad586596.png)

PR #303, opened 2020-02-11
issue #39 - early work on updating the log display

bfitzpat:


bfitzpat:
image

bfitzpat:
And yes, there will be some errors here. Just made sure to get it stashed so we could chat about it.

bfitzpat:
So I now have the webview log working the same as it does in the Kubernetes extension. You can open it and move it around. It suffers from the same problem as the original, not scrolling to the bottom appropriately to follow the log -- so we may need to figure something out there.

Only weirdness is the failing kubectlwatcher tests @apupier  -- I'm not quite sure what those are doing. Any insights into why those would be failing?

bfitzpat:
Another thing we now have to figure out is how best to error out when the pod doesn't exist for a log to follow. Right now we show a note in the Apache Camel K output channel, but that may not be enough.

image

bfitzpat:
And then, when we get it working right, we need to look at how to format it so that errors and warnings are presented nicely in colors other than that drab brown.

bfitzpat:
But we can now launch the Camel K Operator log from the Kubernetes extension views and the Camel K integration logs from the Camel K integrations view and get something like this...

image

bfitzpat:
And need to close the log when we remove the integration. That's another thing left to do.

apupier:
> It suffers from the same problem as the original, not scrolling to the bottom appropriately to follow the log -- so we may need to figure something out there.

this issue https://github.com/Azure/vscode-kubernetes-tools/issues/593 , right?

bfitzpat:
Riccardo gave me the scroll-to-bottom solution, so though that's not the BEST solution it at least updates the log so you can see the latest updates as it scrolls past.

apupier:
please try to provide the fix upstream first to avoid duplicated code that diverge even before the first merge

bfitzpat:
> I guess the provided solution is not covering the issue mentioned int his comment [Azure/vscode-kubernetes-tools#593 (comment)](https://github.com/Azure/vscode-kubernetes-tools/issues/593#issuecomment-516672587) about

You are correct. It does not solve that problem. But it does scroll to the bottom and continuously show the log. I at this point do not know how to turn that off, but I suppose maybe we can add a checkbox to turn it off?

bfitzpat:
> please try to provide the fix upstream first to avoid duplicated code that diverge even before the first merge

I want to comment on this specific line. Though we can contribute things upstream (and I plan to since this was an easy option), since we can't reuse the code directly in VS Code we won't be able to take advantage of that change. And if we change things to add colors for warning/error lines and alter the view further, it's going to diverge very quickly. 

But since the codebases can't be shared due to limitations in extension architecture, our code is going to duplicate upstream effort and diverge regardless.

bfitzpat:
I have added an auto-scroll checkbox to the header on the view that turns that feature on/off and defaults to on. It works pretty well. I think I'm going to remove the filter controls though, since they don't currently do anything. If they get those to work upstream, we can collect that functionality as well -- but until then they just add clutter. 

image

bfitzpat:
Removed the filter fields from our log viewer until they get them working upsream. But here's a screen shot of a Kubernetes-provided log from the camel k operator and one of ours from Camel K. 

image

bfitzpat:
Proposed our solution upstream - https://github.com/Azure/vscode-kubernetes-tools/issues/593#issuecomment-585893634 and https://github.com/Azure/vscode-kubernetes-tools/pull/704

bfitzpat:
I have a weird issue I'm trying to diagnose... If you start a log and then move it to another column/location, we lose the content from the beginning and then it starts again. I suspect it's due to the script in the logsWebview.ts - but I honestly can't figure it out.

apupier:
> > please try to provide the fix upstream first to avoid duplicated code that diverge even before the first merge
> 
> I want to comment on this specific line. Though we can contribute things upstream (and I plan to since this was an easy option), since we can't reuse the code directly in VS Code we won't be able to take advantage of that change. And if we change things to add colors for warning/error lines and alter the view further, it's going to diverge very quickly.
> 
> But since the codebases can't be shared due to limitations in extension architecture, our code is going to duplicate upstream effort and diverge regardless.

if we can keep code duplicated but exactly the same it will be easier to apply fix/improvements done upstream to our codebase. Otherwise it will require potentially conflicting change to backport from upstream to our branch.

for colors for warnings and errors, I think that this is a feature that will be useful upstream too.

Which very specific modifications to Camel K (meaning that won't benefit to upstream too) do you see?

bfitzpat:
> if we can keep code duplicated but exactly the same it will be easier to apply fix/improvements done upstream to our codebase. Otherwise it will require potentially conflicting change to backport from upstream to our branch.

If the code upstream was more mature, I might agree with this. The filter fields in the view have been there in an ornamental state since July 2019. _Duplicating non-working code from upstream and keeping it exactly the same even if it doesn't work doesn't help us._ As a result, we are inevitably going to diverge.

bfitzpat:
Ok, fixed the issue with the log being cut off. Had to do with the two fields for filtering I removed. Enabled those and now it seems to be fine. I think this is a good initial commit for this functionality, which removes the output channel functionality and shifts these logs to individual webview-powered windows that can be moved around the vscode interface. 

We will have to investigate the idea of "pretty print" for the log output, which can likely be handled as we push each log line (after parsing by newlines) to the webview. We can do that separately.

bfitzpat:
> We will have to investigate the idea of "pretty print" for the log output, which can likely be handled as we push each log line (after parsing by newlines) to the webview. We can do that separately.

If we're truly pushing for a duplication of the code from the Kubernetes extension, perhaps log formatting is less important at this point

src/extension.ts
```diff
@@ -417,46 +404,41 @@ export function shareMessageInMainOutputChannel(msg: string) {
...
+			namespace: ns,
+			containers: undefined,
+			containersQueryPath: `.spec`
+		};
+
+		const cresource = `${resource.namespace}/${resource.kindName}`;
+		const panel = LogsPanel.createOrShow('Loading...', cresource);
+
+		let cmd = `logs -f ${podName}`;
+        await kubectlExe.invokeAsync(cmd, undefined, (proc: ChildProcess) => {
```
apupier: indentation issue

src/extension.ts
```diff
@@ -417,46 +404,41 @@ export function shareMessageInMainOutputChannel(msg: string) {
...
+			if (proc && proc.stdout) {
+            	proc.stdout.on('data', (data: string) => {
```
apupier: indentation issue

src/kubectl.ts
```diff
@@ -44,6 +47,62 @@ class KubectlImpl implements Kubectl {
...
+
+async function internalInvokeAsync(command: string, stdin?: string, callback?: (proc: ChildProcess) => void): Promise<ShellResult | undefined> {
+	const bin = await baseKubectlPath();
+	if (bin) {
+		const binpath = bin.trim();
+		const cmd = `${binpath} ${command}`;
+		let sr: ShellResult | undefined;
+		if (stdin) {
+			sr = await exec(cmd, stdin);
+        } else if (callback) {
```
apupier: indentation issue

src/kubectl.ts
```diff
@@ -44,6 +47,62 @@ class KubectlImpl implements Kubectl {
...
+		}
+		return sr;
+	}
+}
+
+async function execStreaming(cmd: string, callback: (proc: ChildProcess) => void): Promise<ShellResult | undefined> {
+    try {
+        return await execCore(cmd, null, callback);
+    } catch (ex) {
+        console.log(ex);
```
apupier: Why not sending to the outputchannel?

src/kubectl.ts
```diff
@@ -44,6 +47,62 @@ class KubectlImpl implements Kubectl {
...
+        return undefined;
+    }
+}
+
+async function exec(cmd: string, stdin?: string): Promise<ShellResult | undefined> {
+    try {
+        return await execCore(cmd, null, null, stdin);
+    } catch (ex) {
+        console.log(ex);
```
apupier: Why not sending to the outputchannel?

src/kubectl.ts
```diff
@@ -44,6 +47,62 @@ class KubectlImpl implements Kubectl {
...
 	}
+
+	async invokeAsync(command: string, stdin?: string, callback?: (proc: ChildProcess) => void): Promise<ShellResult | undefined> {
+		return internalInvokeAsync(command, stdin, callback);
+	}
+	
+}
+
+export interface ShellResult {
+    readonly code: number;
```
apupier: space instead fo tabs used

src/kubectl.ts
```diff
@@ -44,6 +47,62 @@ class KubectlImpl implements Kubectl {
...
+    try {
+        return await execCore(cmd, null, callback);
+    } catch (ex) {
+        console.log(ex);
+        return undefined;
+    }
+}
+
+async function exec(cmd: string, stdin?: string): Promise<ShellResult | undefined> {
+    try {
```
apupier: space instead fo tabs used

src/kubectl.ts
```diff
@@ -44,6 +47,62 @@ class KubectlImpl implements Kubectl {
...
+        return await execCore(cmd, null, null, stdin);
+    } catch (ex) {
+        console.log(ex);
+        return undefined;
+    }
+}
+
+function execCore(cmd: string, opts: any, callback?: ((proc: ChildProcess) => void) | null, stdin?: string): Promise<ShellResult> {
+    return new Promise<ShellResult>((resolve) => {
```
apupier: space instead fo tabs used

src/logsWebview.ts
```diff
@@ -0,0 +1,172 @@
+import * as vscode from 'vscode';
```
apupier: missing file header

src/webpanel.ts
```diff
@@ -0,0 +1,78 @@
+import * as vscode from 'vscode';
```
apupier: missing header

testFixture/.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
```
apupier: i think this shouldn't be commited

src/logsWebview.ts
```diff
@@ -0,0 +1,172 @@
...
+                    ix += step;
+                    setTimeout(fn, 0);
+                }
+                fn();
+              };
+              eval();
+            </script>
+            </body>
+        </html>`;
+    }
```
apupier: wrong indentation level

src/logsWebview.ts
```diff
@@ -0,0 +1,172 @@
...
+}
```
apupier: wrong indentation level

src/logsWebview.ts
```diff
@@ -0,0 +1,172 @@
...
+            this.panel.webview.postMessage({
+                command: 'content',
+                text: content,
+            });
+        }
+    }
+
+    protected update() {
+        if (this.panel.visible) {
+        this.panel.title = `Logs - ${this.resource}`;
```
apupier: wrong indetantion level

apupier:
> > We will have to investigate the idea of "pretty print" for the log output, which can likely be handled as we push each log line (after parsing by newlines) to the webview. We can do that separately.
> 
> If we're truly pushing for a duplication of the code from the Kubernetes extension, perhaps log formatting is less important at this point

i think we should provide the feature upstream first. this is not a specific feature for Camel K needs

lhein:
> most of the commit messages are missing the issue number

I think we can do a squash commit with a corrected issue number. Its just 8 files and not worth the hassle to rewrite the commit history.

src/kubectl.ts
bfitzpat: Fixed

src/kubectl.ts
bfitzpat: Fixed

src/webpanel.ts
bfitzpat: Added MIT license header and mention that we got this from the kubernetes tooling project

src/extension.ts
```diff
@@ -100,10 +98,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
 				}).catch( (err) => {
 					console.log(err);
 				});
-				await removeOutputChannelForIntegrationViaKubectl(integrationName)
-				.catch( (err) => {
-					console.log(err);
-				});
+				// TODO: we need to look into closing the log view when the integration is stopped
```
apupier: have you created a specific issue yet?

src/logsWebview.ts
apupier: there is still a wrong indentation level. The indentation is correct upstream https://github.com/Azure/vscode-kubernetes-tools/blob/a7d45e92e39036256c610cec9cb0245c96453fb2/src/components/logs/logsWebview.ts#L161

apupier:
> > most of the commit messages are missing the issue number
> 
> I think we can do a squash commit with a corrected issue number. Its just 8 files and not worth the hassle to rewrite the commit history.

this is an hassle of 20-30 seconds for a simple squash... which allows to review the commit message and history. But already discussed directly with Brian which doesn't see the value either.

src/extension.ts
lhein: https://github.com/camel-tooling/vscode-camelk/issues/310

lhein: I also created https://github.com/camel-tooling/vscode-camelk/issues/311 to track the issue of the stale log views when updating an integration changes the pod.

lhein:
@apupier I think we can agree on doing a squash into a single commit and force push that one with a correct issue number and title?

apupier:
> @apupier I think we can agree on doing a squash into a single commit and force push that one with a correct issue number and title?

this is the minimal solution that I proposed.

README.md
```diff
@@ -218,13 +218,13 @@ If you want to explicitly view the running log for a published integration in a
...
 ![Apache Camel K Integrations view - Follow log](images/camelk-integrations-view-remove-menu.jpg)
 
-A new Output channel opens. It is named for the running Kubernetes pod where the integration is running. This log updates as new data is added:
+A new view opens. It is named for the running Kubernetes pod where the integration is running. This log updates as new data is added:
 
-![Apache Camel K Integrations view - Integration log](images/camelk-integrations-view-integrations-log.jpg)
+![Apache Camel K Integrations view - Integration log](images/camelk-integrations-view-integrations-log-revised.jpg)
 
-## Stopping an Apache Camel K integration
+Note: By default, `Auto-scroll` is checked and new entries in the log will automatically be scrolled to at the bottom of the window. If turned off, the scrolling stops and can be done manually.
```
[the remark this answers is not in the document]
bfitzpat: I disagree. The sentence makes sense as is. "new entries in the log will automatically be scrolled to at the bottom of the window". Leaving it.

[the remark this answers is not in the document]
lhein: You are the native speaker. Fine for me.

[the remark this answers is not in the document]
apupier: fine for me too. you're the native speaker.
I reported because it feels strange to have two prepositions and Google doc autocorrection is suggesting to remove the 'at'

src/logsWebview.ts
apupier: this is fine now (I should have put my comment on the line above so that it would have marked my comments as outdated automatically)

PR #312, opened 2020-02-19
FUSETOOLS2-224 - provide template for Camel K in tasks.json

apupier:
- single static template for starting
- using CompletionItemprovider, it is the same API than for a Language
Server, so the day we want to extract it as a more generic tool it
should be relatively easy
- using a jsonc-parser. it is the parser used by
vscode-json-languageservice. The VS Code JSOn Language Service is not
proposing useful API to reuse directly.
- the computation of the place where to provide the completion is a bit
handcrafted, would be best to rely on json schema in the future.
Currently, it will provide false-positive only if a tasks.json is used
and is not a VS Code Tasks files and that inside there is a tasks
attribute with an Array. I think the risk of false positive is low
enough for a first iteration.

completionTasksJson


bfitzpat:
I really like this approach (after a rebase). How easy will it be to add other templates going forward?

apupier:
> How easy will it be to add other templates going forward?

for templates that doesn't require access to extra environment context, just need to add a json from a string in the completion array: https://github.com/camel-tooling/vscode-camelk/pull/312/files#diff-e41d2ed4337ccccb89f8ab480b6b65aeR30-R43

We can have access to the context through vscode.CompletionContext https://github.com/camel-tooling/vscode-camelk/pull/312/files#diff-e41d2ed4337ccccb89f8ab480b6b65aeR24 (for instance to propose a template for all Camel K files detected in the workspace dynamically).
So not complicated, will just require a bit more coding but that's normal it is a more advanced use case.

apupier:
> Should we add something to the readme?

I was wondering too. I felt that there is no real need as it is available in normal completion. Maybe will be worthier when there will be more completion available? including some dynamic one and for some attributes values

bfitzpat:
Let's focus on making the functionality (which is awesome) more available to the user beyond just the existing task commands in VS Code. That will likely be more worthy of the readme I think. :) No worries.

Issue #311, opened 2020-02-19
log view can become stale on integration updates


lhein:
if the pod changes on an update of the integration then the old log view becomes stale and a new one needs to be opened which has a different title.

bfitzpat:
Looks like we can use the kamel cli to follow logs so this doesn't occur. 

i.e. `kamel log [myintegration]` will update the pod when you redeploy the integration and continues scrolling the log

bfitzpat:
Created https://github.com/apache/camel-k/issues/1283

bfitzpat:
Created https://github.com/camel-tooling/vscode-camelk/pull/313

djelinek:
verified - redhat.vscode-camelk 0.0.14-413 _(snapshot)_, VS Code 1.43.2

PR #313, opened 2020-02-19
issue #311 - shifting to kamel log instead of kubectl log

bfitzpat:

### DIFF
--- a/src/kubectlutils.ts
+++ b/src/kubectlutils.ts
@@ -180,3 +180,24 @@ export async function getKubernetesVersion(): Promise<string | undefined> {
 
     return `${major}.${minor}`;
 }
+
+// retrieve the first integration running in camel k starting with the integration name
+export async function getIntegrationsFromKubectl(integrationName : string): Promise<string> {
+	return new Promise <string> ( async (resolve, reject) => {
+		await getPodsFromKubectlCli()
+		.then( (allPods) => {
+			let podArray = parseShellResult(allPods);
+			podArray.forEach(podName => {
+				if (podName.startsWith(integrationName)) {
+					resolve(podName);
+					return;
+				}
+			});
+			resolve(undefined);
+			return;
+		}).catch ( (error) => {
+			reject(error);
+			return;
+		});
+	});
+}
--- a/src/kubectlutils.ts
+++ b/src/kubectlutils.ts
@@ -180,24 +180,3 @@ export async function getKubernetesVersion(): Promise<string | undefined> {
 
     return `${major}.${minor}`;
 }
-
-// retrieve the first integration running in camel k starting with the integration name
-export async function getIntegrationsFromKubectl(integrationName : string): Promise<string> {
-	return new Promise <string> ( async (resolve, reject) => {
-		await getPodsFromKubectlCli()
-		.then( (allPods) => {
-			let podArray = parseShellResult(allPods);
-			podArray.forEach(podName => {
-				if (podName.startsWith(integrationName)) {
-					resolve(podName);
-					return;
-				}
-			});
-			resolve(undefined);
-			return;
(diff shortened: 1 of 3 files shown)


bfitzpat:
Switching to `kamel log [integration]` instead of `kubectl logs-f [podname]` gives us uninterrupted coverage of the running integration, including when the integration is re-deployed with changes. The issue was that using kubectl required us to follow a single pod, which is then discarded when an integration is re-published. The `kamel` version follows the integration through shutting down a running pod and spinning a new one up.

The only problem I have found is a complete lack of log feedback when the kit is created and the pod is first spinning up. There is just... nothing. Without any kind of user feedback, some users may wonder if the integration will ever spin up.

I have asked for an update in https://github.com/apache/camel-k/issues/1283 so that the user can see something moving while waiting for the integration to start. Running `kamel log [integration]` at the command line has the same problem.

src/kubectlutils.ts
```diff
@@ -180,3 +180,24 @@ export async function getKubernetesVersion(): Promise<string | undefined> {
...
 }
+
+// retrieve the first integration running in camel k starting with the integration name
+export async function getIntegrationsFromKubectl(integrationName : string): Promise<string> {
+	return new Promise <string> ( async (resolve, reject) => {
+		await getPodsFromKubectlCli()
+		.then( (allPods) => {
+			let podArray = parseShellResult(allPods);
+			podArray.forEach(podName => {
+				if (podName.startsWith(integrationName)) {
```
apupier: Using the first one matching the substring can lead to using the wrong pod, isn't it? Do you know the exact case it can happen?

some other thoughts:
Has the suffix a fixed length? In this case we can substring. I think it would be worthy to handle in this PR in this case (unless we already had the same issue before?)
Otherwise, we might have false positive, maybe we should detect when several are matching and:
- log it as a first iteration in this PR
- propose user to choose the one he wants? This one seems more complicated and might be easier to handle in a separate PR.
- or have another heuristic which is deterministic?

src/logsWebview.ts
```diff
@@ -106,7 +106,7 @@ export class LogsPanel extends WebPanel {
 						switch (message.command) {
 							case 'content':
 								const elt = document.getElementById('content');
-								const text = message.text.split('\\n');
+								const text = message.toString().split('\\n');
```
apupier: what is the advantage of toString() versus getting the "text"? or is it that the retrieved message doesn't have the same structure when using kamel cli instead of kubectl?

src/kubectlutils.ts
bfitzpat: Couple of things.

1. This is utility code we have been using for a while to get the pod and retrieve the log for the pod that we used in the output channel. We are not using it any longer (now using kamel log instead of kubectl --logs) but I wanted to keep it in case we needed it in the future. 
2. Because of #1, maybe we should just remove it entirely. 

But to be honest, it has worked great for months with no issues. We're just shifting to a different method of tracking the logs now and don't really need it.

src/logsWebview.ts
bfitzpat: I was hitting a problem in the hosted javascript where it was throwing an error stating that split was not available on message.text. This was the workaround. I suspect what you are saying is true - that there's a difference in the structure of the message returned.

src/kubectlutils.ts
bfitzpat: And yes, I wish we could test - but without a solid way of integration testing this thing I don't know quite what to do on that front.

apupier: > 1. but I wanted to keep it in case we needed it in the future.

if we remove it, we will still have it in the history :-)

> 2\. Because of #1, maybe we should just remove it entirely.

I vote for removing it :-)

Issue #310, opened 2020-02-19
close the log view when the integration is stopped


lhein:

bfitzpat:
I'm wondering if this should be optional. Are there cases where waiting to see if a camel log shuts down without errors is a benefit? Especially with https://github.com/camel-tooling/vscode-camelk/issues/311 fixed, there's a lot more in that log around watching integrations spin up and shut down when they are updated.

apupier:
agree.
Maybe having a marker to show that the integration corresponding to the log has been stopped would be enough? Italic title? a suffix in the title such as [stopped]?

we can still imagine a settings to automatically close the view? Either one in VS Code or a checkbox on the log view?

bfitzpat:
I will see if we can watch for the items in the log to appear and update the title of the window accordingly. That's a good idea.

It appears that we may be able to simply dispose the panel (i.e. `Extensions can also programmatically close webviews by calling dispose() on them` from https://code.visualstudio.com/api/extension-guides/webview#lifecycle) that I hadn't seen before. I will see what we can do to leverage that and add a setting.

bfitzpat:
Addressed both issues with https://github.com/camel-tooling/vscode-camelk/pull/316

djelinek:
verified - redhat.vscode-camelk 0.0.14-413 _(snapshot)_, VS Code 1.43.2

PR #316, opened 2020-02-20
issue #310 - working on closing the log view when the integration is …

bfitzpat:
…stopped


bfitzpat:
image

bfitzpat:
Added `[Integration stopped]` in the title when the integration is removed. 

Starting to work on the new preference and shutting down the view if the preference is set to true.

bfitzpat:
image

bfitzpat:
Code now removes the logview if the remove setting is set to true.

bfitzpat:
> when removing an integration which has not finished to start with settings set to "not close", the log view title is not updated

I honestly don't understand what you're saying here. Can you clarify? 

I THINK you're saying that if you:
* start a new integration (without a previously running kit, which means it takes a bit to spin up) 
* open the log (which sometimes takes a long time to populate with anything from the integration)
* remove the integration (while it's still coming up)
* the log title doesn't change. 

This is pretty easy to explain. Right now I'm looking for the text "Received hang up - stopping the main instance" in the log itself. If that doesn't show up, I don't know to change the title.

Seems we should also change the panel title if we remove the integration and there's no actual content.

bfitzpat:
Updated the code to handle the two missing cases

src/extension.ts
```diff
@@ -368,3 +386,36 @@ function handleLogViaKamelCli(integrationName: string) : Promise<string> {
 export function getStashedContext() : vscode.ExtensionContext {
 	return stashedContext;
 }
+
+function removeIntegrationLogView(integrationName: string) : Promise<string> {
+	return new Promise<string>( async () => {
+		if (closeLogViewWhenIntegrationRemoved) {
+			LogsPanel.currentPanels.forEach((value : LogsPanel) => {
+				var title = value.getTitle();
+				if (title.indexOf(integrationName) > 0) {
```
apupier: might be better to check >= 0, in case the integration name is put at te beginning of the title one day

src/extension.ts
```diff
@@ -368,3 +386,36 @@ function handleLogViaKamelCli(integrationName: string) : Promise<string> {
...
+					value.disposeView();
+				}
+			});
+		} else {
+			LogsPanel.currentPanels.forEach((value : LogsPanel) => {
+				var title = value.getTitle();
+				if (title.indexOf(integrationName) > 0) {
```
apupier: might be better to check >= 0, in case the integration name is put at te beginning of the title one day

src/extension.ts
```diff
@@ -368,3 +386,36 @@ function handleLogViaKamelCli(integrationName: string) : Promise<string> {
...
+			});
+		}
+	});
+}
+
+function updateLogViewTitleToStopped(panel: LogsPanel, title: string) {
+	if (panel && title) {
+		// make sure we only show the log as stopped once 
+		const stoppedString = ` [Integration stopped]`;
+		let boolHasIntegrationStopped = title.indexOf(stoppedString) > 0;
```
apupier: might be better to check >= 0, in case the "[Integration stopped]" is put at te beginning of the title one day, and check without the first space in this case

Issue #321, opened 2020-02-24
no easy way to open the operator log or logs for individual integration kits


bfitzpat:
Two of the the logs that would be very helpful when debugging integration issues would be the Camel K Operator log and any log associated with a particular Camel K integration kit. Though you can get to them through the Kubernetes view, it would be nice if we could provide a short cut to get to them via the Integrations view.

bfitzpat:
Adding a new command button to the integrations view with the Note icon from https://github.com/microsoft/vscode-icons 
![image](https://user-images.githubusercontent.com/530878/75190143-ea478080-570c-11ea-974b-21283c4ce2a1.png)

![image](https://user-images.githubusercontent.com/530878/75190320-49a59080-570d-11ea-86b8-e5ad769b0fc5.png)

djelinek:
verified - redhat.vscode-camelk 0.0.14-413 _(snapshot)_, VS Code 1.43.2

PR #322, opened 2020-02-24
issue #321 - adding new button on integration view to open camel k op…

bfitzpat:
…erator log


src/extension.ts
```diff
@@ -134,6 +134,43 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
 		// add commands to create config-map and secret objects from .properties files
 		configmapsandsecrets.registerCommands();
 		
+		// create the integration view action -- open operator log
+		vscode.commands.registerCommand('camelk.integrations.openOperatorLog', async () => {
+			setStatusLineMessageAndShow(`Retrieving log for Apache Camel K Operator...`);
+			await kubectlutils.getNamedPodsFromKubectl(`camel-k-operator`).then( async (podNames) => {
+				await handleLogViaKubectlCli(podNames[0]).catch((error) => {
```
apupier: why is it taking the first one? What happens if there are several?

src/extension.ts
```diff
@@ -134,6 +134,43 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+		// create the context menu integration view action -- open kit log
+		vscode.commands.registerCommand('camelk.integrations.kitlog', async (node: TreeNode) => {
+			if (node && node.label) {
+				let integrationName : string = node.label;
+				setStatusLineMessageAndShow(`Retrieving log for Apache Camel K Integration kit...`);
+				await getIntegrationsListFromKamel(integrationName).then( async (result : string ) => {
+					if (result && result.length > 0) {
+						let lines : string[] = result.split("\n");
+						if (lines.length > 1) {
+							let secondLine = lines[1]; // "integration\tState\tkit-pod-name"
```
apupier: what means the comment?

bfitzpat:
@apupier You realize that this was "work in progress" and not "ready for review", right?

src/extension.ts
bfitzpat: there should only be one for a given namespace

bfitzpat: but I am asking the runtime guys to verify

src/extension.ts
bfitzpat: "work in progress" is just that. :)

src/extension.ts
```diff
@@ -134,6 +134,43 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
...
+				await handleLogViaKubectlCli(podNames[0]).catch((error) => {
+					utils.shareMessage(mainOutputChannel, `error: ${error} \n`);
+				});
+			}).catch( (err) => {
+				console.log(err);
+			});
+			hideStatusLine();
+		});
+		// create the context menu integration view action -- open kit log
+		vscode.commands.registerCommand('camelk.integrations.kitlog', async (node: TreeNode) => {
```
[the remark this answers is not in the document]
bfitzpat: yes, definitely some clean-up to be done moving things around

src/extension.ts
bfitzpat: Seems based on info from the runtime folks that there will be at most 1, but may be none.

bfitzpat:
>     * when removing the integration, the kit builder log should follow the same behavior than "normal" log. I.e. automatically close or put [Integration stopped]

That is incorrect. The kit builders will continue to run in the background until a `kamel reset` is called to clean them up. They do not get disposed. 

>     * what about adding a test checking that fro a given result of the cli, it can parse it and call the right function? it will help validating that the parsing is correct. it will allow to better understand what is expected output of the CLI.

I'm guessing you're talking about the kamel cli information parsing when I do a `kamel get` and parse to get the kit name? Can you clarify what you're suggesting here @apupier ?

apupier:
I think would be more for another iteration, what do you think to separate the "remove" from the 2 "logs" contextual menu with a line?

bfitzpat:
> I think would be more for another iteration, what do you think to separate the "remove" from the 2 "logs" contextual menu with a line?

We can do it, but the line is impossible to see regardless of the color theme

image

image

Not sure it's worth it.

bfitzpat:
> * not important: the screenshots contains a "typo" for the workspace name: "camelk-exmples". If you're very courageous, you can change it otherwise that's fine with that

Can you tell me which image has the misspelling in it? 

> * optional (can be another PR): when clicking on "Open Apache Camel K Operator log" although the deployment has not been done yet. The error message is starting with 2 "Error" word. `Error Error from server (BadRequest): container "camel-k-operator" in pod "camel-k-operator-6bcf77ddf8-llmp4" is waiting to start: ContainerCreating`

Yes, I noticed that as well. Definitely good for a separate issue. https://github.com/camel-tooling/vscode-camelk/issues/325

> * works great, small little feedback inlined

Thanks.

bfitzpat:
@apupier Corrections made except for double error error issue and menu divider, which we will investigate separately

apupier:
> > I think would be more for another iteration, what do you think to separate the "remove" from the 2 "logs" contextual menu with a line?
> 
> We can do it, but the line is impossible to see regardless of the color theme
> 
> image
> 
> image
> 
> Not sure it's worth it.

I think there is a bug with VS Code on your system.

I can see the lines correctly:
Screenshot from 2020-02-26 15-11-11

bfitzpat:
> I think there is a bug with VS Code on your system.

Could be an issue with this variety of linux. We can look into it later, I don't think it's critical for this PR.

apupier:
> > I think there is a bug with VS Code on your system.
> 
> Could be an issue with this variety of linux. We can look into it later, I don't think it's critical for this PR.

ok, created https://issues.redhat.com/browse/FUSETOOLS2-235

Issue #280, opened 2020-01-27
Provide user a way to circumvent the auto-upgrade of Kamel CLI version installation


bfitzpat:
With https://github.com/camel-tooling/vscode-camelk/pull/274, it was discussed that that PR ensures that every time the extension activates, it updates the kamel CLI to the latest version as specified in the installer.ts file (https://github.com/camel-tooling/vscode-camelk/blob/146d02527cba00469fd717497f234eb82b10285e/src/installer.ts#L34). It does not ask, it just updates the cli. 

This is fine as the default until we come to cases where the latest CLI may not work on a particular platform. Perhaps it won't quite work on Che as expected with a particular version and needs to be downgraded to an older one. That should be a valid case. 

As I point out in the PR, the only thing I can think is to offer another preference that essentially says "turn off automatic update of CLI" and works around the version check. If the user checks the box (which is unchecked by default), the installer avoids the auto-update.

bfitzpat:
Another possibility is to raise the preference -- so the user can explicitly specify which version of the CLI to install. Perhaps an editable field that shows the version currently used (defaults to the one specified as noted above) and will attempt to download the version the user specifies if it is overridden.

bfitzpat:
Created https://issues.redhat.com/browse/FUSETOOLS2-208 to track this in JIRA as well

bfitzpat:
Adding an "Autoupdate" checkbox. This does a couple of things to keep the version in check.

1. The version will always update to the latest default value set in the extension (right now 1.0.0-RC2)
2. If the user changes the version, but leaves autoupdate checked, it will reset it to the latest default.
3. If the user turns off autoupdate and then changes the version, the version they set sticks.
4. If they then turn on autoupdate, the version goes back to the default (1.0.0-RC2)

![image](https://user-images.githubusercontent.com/530878/75724174-09a95500-5c9b-11ea-9e86-987b4eb7c4bc.png)

Note that the user must restart the workspace (and they get a little info popup saying that the version has changed and they need to restart) to get the version of the runtime they specify. 

Also note that there's a chance that the Camel K CLI version can get out of whack with what is installed in a particular runtime (Kubernetes, OpenShift). At this time there's no way for the user to see the difference, but perhaps that's something we should look into?

Also note that we don't verify that the version is available currently, so that should happen as well.

bfitzpat:
[...]

PR #329, opened 2020-02-28
FUSETOOLS2-239 - update to runtime 1.0.0-RC2

bfitzpat:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to the "vscode-camelk" extension will be documented in this 
 - Add an extension preference to close an open log window if the integration is removed
 - Add a new button to the Integrations view that opens the Apache Camel K Operator log
 - Add a new context menu to published integrations in the Integrations view to open the Kit Builder log
+- Update default runtime version to 1.0.0-RC2
 
 ## 0.0.12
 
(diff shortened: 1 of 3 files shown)


bfitzpat:
> the version of the Camel artefact to download must be updated to the version used by this new version of Camel K

It doesn't. We pick up the latest Camel K tag right now in https://github.com/camel-tooling/vscode-camelk/blob/48e33f9876fd7172800aed2c1668d7553d192e91/src/installer.ts#L221

That comes from https://api.github.com/repos/apache/camel-k/releases/latest - which provides the correct tag name 1.0.0-RC2.

bfitzpat:
We still need a way to circumvent that -- as recorded in https://github.com/camel-tooling/vscode-camelk/issues/280 -- so I'll work on that next

apupier:
I was mentioning this version https://github.com/camel-tooling/vscode-camelk/blob/48e33f9876fd7172800aed2c1668d7553d192e91/src/JavaDependenciesManager.ts#L28 which needs to be aligned to this one https://github.com/apache/camel-k-runtime/blob/89b79bd1c347509b407f9454174ace4f4b12f6f6/pom.xml#L38

bfitzpat:
> version for Java completion needs to be updated to 3.0.1

Thanks for the clarification. Just updated that.

PR #330, opened 2020-02-28
issue #280 - first cut of setting to avoid auto-upgrade of kamel cli

bfitzpat:

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -22,7 +22,7 @@ import * as utils from './CamelKJSONUtils';
 import * as configmapsandsecrets from './ConfigMapAndSecrets';
 import * as integrationutils from './IntegrationUtils';
 import * as events from 'events';
-import { installKamel, installKubectl, checkKamelNeedsUpdate } from './installer';
+import { installKamel, installKubectl} from './installer';
 import { Errorable, failed } from './errorable';
 import * as kubectl from './kubectl';
 import * as kamel from './kamel';
@@ -34,6 +34,7 @@ import { CamelKTaskProvider } from './task/CamelKTaskDefinition';
 import { ChildProcess } from 'child_process';
 import { LogsPanel } from './logsWebview';
 import * as logUtils from './logUtils';
+import {checkKamelNeedsUpdate, version, handleChangeRuntimeConfiguration} from './versionUtils';
 
 export const DELAY_RETRY_KUBECTL_CONNECTION = 1000;
 
@@ -48,6 +49,7 @@ const restartKubectlWatchEvent = 'restartKubectlWatch';
 let runningKubectl : ChildProcess | undefined;
 let timestampLastkubectlIntegrationStart = 0;
 export let closeLogViewWhenIntegrationRemoved : boolean;
+export var runtimeVersionSetting : string | undefined;
 
 let stashedContext : vscode.ExtensionContext;
 
@@ -286,6 +288,32 @@ function applyLogviewSettings(): void {
 	});
 }
 
+async function applyDefaultVersionSettings() : Promise<void> {
+	let runtimeSetting = config.getKamelRuntimeVersionConfig();
+	let autoUpgradeSetting = config.getKamelAutoupgradeConfig();
+	if (!runtimeSetting) {
+		await config.setKamelRuntimeVersionConfig(version).then ( () => {
+			runtimeVersionSetting = version;
+			shareMessageInMainOutputChannel(`No version found. Using default version ${runtimeVersionSetting} of Apache Camel K CLI`);
+		});
+	} else {
+		if (autoUpgradeSetting && autoUpgradeSetting === true) {
+			if (runtimeSetting && version.toLowerCase() !== runtimeSetting.toLowerCase()) {
+				await config.setKamelRuntimeVersionConfig(version).then ( () => {
+					runtimeVersionSetting = version;	
+					shareMessageInMainOutputChannel(`Auto-upgrade setting enabled. Updating to default version ${version} of Apache Camel K CLI`);
+				});
+			}
+		} else {
+			shareMessageInMainOutputChannel(`Using version ${runtimeSetting} of Apache Camel K CLI`);
+			runtimeVersionSetting = runtimeSetting;
+		}
+	}
+	vscode.workspace.onDidChangeConfiguration(async () => {
+		await handleChangeRuntimeConfiguration();
+	});
+}
+
 function refreshIfNamespaceChanges(): void {
 	vscode.workspace.onDidChangeConfiguration(async () => {
 		if (camelKIntegrationsTreeView && camelKIntegrationsTreeView.visible === true) {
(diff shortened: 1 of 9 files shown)


bfitzpat:
@lhein not sure this is the best approach. Any thoughts?

src/installer.ts
```diff
@@ -148,6 +151,11 @@ export async function checkKamelNeedsUpdate(versionToUse? : string): Promise<boo
 			}
 			versionToUse = latestversion.result.trim();
 		}
+		if (runtimeVersionSetting && runtimeVersionSetting.toLowerCase() !== versionToUse.toLowerCase()) {
+			versionToUse = runtimeVersionSetting;
```
apupier: I think the the failing test is reporting a real problem. The version to use is modified here from "bogusversion" to "1.0.0-RC2", then the method declares that there is no kamel update needed.

src/extension.ts
```diff
@@ -286,6 +287,48 @@ function applyLogviewSettings(): void {
...
+				});
+			}
+		} else if (runtimeVersionSetting && runtimeVersionSetting.toLowerCase() !== runtimeSetting.toLowerCase() && newUpgradeSetting === false) {
+			const msg = `Setting for default Apache Camel K runtime version has changed. Please restart the workspace to refresh the Camel K cli.`;
+			shareMessageInMainOutputChannel(msg);
+			vscode.window.showWarningMessage(msg);
+			runtimeVersionSetting = runtimeSetting;
+		}
+	});
+}
```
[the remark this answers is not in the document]
bfitzpat: If it was more than a couple of functions, I might agree. But at this point there's not enough here to warrant moving it to a separate file.

src/installer.ts
bfitzpat: You are correct - this is what launched me into looking at why I can't run tests locally (see https://github.com/camel-tooling/vscode-camelk/pull/335). Can't debug a test if I can't run ANY tests.

src/extension.ts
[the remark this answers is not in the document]
bfitzpat: After working the test a bit and reworking some of the functions, I moved them into a specific version utility file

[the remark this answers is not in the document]
bfitzpat: But I left the setting changes in extension for now. Couldn't find a clean way to extricate the code.

bfitzpat:
Huzzah! Fixed all tests locally and up on CI. 43 tests succeeding here and on travis. I think we have a winner.

src/extension.ts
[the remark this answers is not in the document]
bfitzpat: Have moved those now.

src/versionUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+	return new Promise<boolean>( async (resolve, reject) => {
+		let runtimeVersionSetting = vscode.workspace.getConfiguration().get(config.RUNTIME_VERSION_KEY) as string;
+
+		if (versionToUse) {
+			let thisVersionAvailable = await testVersionAvailable(versionToUse);
+			if (!thisVersionAvailable) {
+				var msg = `Camel K CLI Version ${versionToUse} unavailable.`;
+                extension.shareMessageInMainOutputChannel(msg);
+                reject(new Error(msg));
+                return false;
```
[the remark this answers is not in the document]
bfitzpat: I believe I have all the duplicated messages taken care of now.

bfitzpat:
@apupier you'll notice I've added some tests. unfortunately, the runtime version test fails. I left it in with "skip" so you could see the idea there - but for some reason, it's not working. I'm open to ideas.

apupier:
> unfortunately, the runtime version test fails. I left it in with "skip" so you could see the idea there - but for some reason, it's not working. I'm open to ideas.

in the local folder testFixture, there is a .vscode folder which is not part of pushed source code (and it is great). This is the workspace settings file used when launching tests. Due to previous run, there is the property set of the Camel K runtime version.
With new code, when updating value of this property programmatically, it is modifying the value in Global settings. When we try to do a get, the workspace settings previously defined is overriding the global version.
Removing the Camle K runtime version settings from your local filesystem will make the test passing.

apupier:
I updated the branch to enable the test. it is passing well on CI as it is a fresh test workspace :-)

src/versionUtils.ts
```diff
@@ -0,0 +1,159 @@
...
+}
+
+export async function handleChangeRuntimeConfiguration() {
+	let runtimeSetting = config.getKamelRuntimeVersionConfig();
+	let newUpgradeSetting = config.getKamelAutoupgradeConfig();
+	if (newUpgradeSetting === true) {
+		if (runtimeSetting && version.toLowerCase() !== runtimeSetting.toLowerCase()) {
+			extension.shareMessageInMainOutputChannel(`Auto-upgrade setting enabled. Updating to default version ${version} of Apache Camel K CLI`);
+			await config.setKamelRuntimeVersionConfig(version).then ( () => {
+				const msg = `Setting for Apache Camel K runtime version has changed to default ${version}. Please restart the workspace to refresh the Camel K cli.`;
```
apupier: why do I need to restart the workspace? Why just replacing the kamel binary is not enough?

src/versionUtils.ts
```diff
@@ -0,0 +1,159 @@
...
+			const msg = `Setting for Apache Camel K runtime version has changed to ${runtimeSetting}. Please restart the workspace to refresh the Camel K cli.`;
+			extension.shareMessageInMainOutputChannel(msg);
+			vscode.window.showWarningMessage(msg);
+			extension.setRuntimeVersionSetting(runtimeSetting);	
+		}
+	} else if (runtimeSetting && runtimeSetting.trim().length > 0 && version.toLowerCase() !== runtimeSetting.toLowerCase()) {
+		const msg = `Setting for Apache Camel K runtime version has changed to ${runtimeSetting}. Please restart the workspace to refresh the Camel K cli.`;
+		extension.shareMessageInMainOutputChannel(msg);
+		vscode.window.showWarningMessage(msg);
+		extension.setRuntimeVersionSetting(runtimeSetting);
```
apupier: I have hard time to understand this piece of code.
Can you provide more specific function for if statement or if-block please?

Also, i the think the 3 lines can be factorized		```extension.shareMessageInMainOutputChannel(msg);
		vscode.window.showWarningMessage(msg);
		extension.setRuntimeVersionSetting(runtimeSetting);```

src/versionUtils.ts
bfitzpat: I am working to stick with the Kubernetes approach to installing CLIs here, which means doing it once - at startup - and not afterwards. If we want to change that approach, I think we can - but I think it works the way it is and we can address this in a second PR.

src/versionUtils.ts
bfitzpat: Will see what I can do

bfitzpat: Added docs and refactored with suggestion

apupier: The comments are helping to understand the code even if having refactored code with clear method names would have been better as it would ensure a better maintainability.

PR #356, opened 2020-03-23
FUSETOOLS2-139 - Command to create a new Apache Camel K integration file

apupier:
- requires a root folder
- can be created only at the root
- only create the file, doesn't open it due to issues with kamel.exec
not returning. So will be improved in another iteration, see https://issues.redhat.com/browse/FUSETOOLS2-303

NewApacheCamelKFileCreation

please note that a kubeconfigf needs to be available currently, see [here](https://github.com/apache/camel-k/issues/1368)
Potential error on creation are not reported but it is linked to the problem of kamel.exec

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,15 @@ The **Tooling for Apache Camel K** extension is available in the VS Code Extensi
 
     You can install them by following the same steps except search for Kubernetes or Apache Camel in the list of extensions.
 
+## Creating a new Camel K Integration file
+
+1. From command palette, choose `Create a new Apache Camel K Integration file`
+2. Choose the language to use.
+3. Choose the workspace folder
+4. Provide a name for the new file
+
+The new Apache Camel K Integration file is created at the root of the chosen workspace folder.
+
 ## Starting a new Camel K Integration
 
 After your Apache Camel K/Minikube environment is running and you have installed the **Tooling for Apache Camel K** (vscode-camelk) extension, you can start a new Apache Camel K integration.
@@ -106,7 +115,7 @@ To update the state of your currently deployed integrations, hover over the **Ap
 
 ![Apache Camel K Integrations view - Refresh](images/camelk-integrations-view-refresh-action.jpg)
 
-## Creating a new Camel K Integration with multiple parameters
+## Creating a new Camel K Integration task configuration with multiple parameters
 
 Though the simple "Start Apache Camel K Integration" menu works well for simple cases, you can use a Task for more complex integrations. When the Camel K integration requires more configuration, you can set that up using a Task.
 
--- a/src/commands/NewIntegrationFileCommand.ts
+++ b/src/commands/NewIntegrationFileCommand.ts
@@ -33,9 +33,10 @@ const LANGUAGES_WITH_FILENAME_EXTENSIONS = new Map([
 const LANGUAGES = Array.from(LANGUAGES_WITH_FILENAME_EXTENSIONS.keys());
 
 export async function create() : Promise<void> {
-	const language = await vscode.window.showQuickPick(LANGUAGES);
+	const language = await vscode.window.showQuickPick(LANGUAGES, {placeHolder:'Please select the language in which the new file will be generated.'});
 	if (language) {
-		const workspaceFolder = await vscode.window.showWorkspaceFolderPick();
+		const workspaceFolder = await vscode.window.showWorkspaceFolderPick(
+			{placeHolder: 'Please select the workspace folder in which the new file will be created.'});
 		if (workspaceFolder) {
 			const filename = await vscode.window.showInputBox({
 				prompt: 'Please provide a name for the new file (without extension)',
--- a/src/commands/NewIntegrationFileCommand.ts
+++ b/src/commands/NewIntegrationFileCommand.ts
@@ -1,0 +1,85 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+import * as fs from 'fs';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import * as kamel from '../kamel';
+
+const validFilename = require('valid-filename');
+
+const LANGUAGES_WITH_FILENAME_EXTENSIONS = new Map([
+	['Java', 'java'],
+	['XML', 'xml'],
+	['Yaml', 'yaml'],
+	['Groovy', 'groovy'],
+	['JavaScript', 'js'],
+	['Kotlin', 'kts']]);
+const LANGUAGES = Array.from(LANGUAGES_WITH_FILENAME_EXTENSIONS.keys());
+
+export async function create() : Promise<void> {
+	const language = await vscode.window.showQuickPick(LANGUAGES);
+	if (language) {
+		const workspaceFolder = await vscode.window.showWorkspaceFolderPick();
+		if (workspaceFolder) {
+			const filename = await vscode.window.showInputBox({
+				prompt: 'Please provide a name for the new file (without extension)',
+				validateInput: (name: string) => {
+					return validateFileName(name, language, workspaceFolder);
+				}
+			});
+			if (filename) {
+				const kamelExe = kamel.create();
+				const newFileFullPath: string = computeFullpath(language, workspaceFolder, filename);
+				kamelExe.invoke(`init ${newFileFullPath}`);
+			}
+		}
+	}
+}
+
+function computeFullpath(language: string, workspaceFolder: vscode.WorkspaceFolder, filename: string): string {
+	const extension = getFileExtensionForLanguage(language);
+	return path.join(workspaceFolder.uri.fsPath, `${filename}.${extension}`);
+}
+
+export function validateFileName(name: string, language: string, workspaceFolder: vscode.WorkspaceFolder): string | undefined {
+	if (!name) {
+		return 'Please provide a name for the new file (without extension)';
+	}
+	let newFilePotentialFullPath: string = computeFullpath(language, workspaceFolder, name);
+	if (fs.existsSync(newFilePotentialFullPath)) {
+		return 'There is already a file with the same name. Please choose a different name.';
+	}
+	if (!validFilename(name)) {
+		return 'The filename is invalid.';
+	}
+	const patternJavaNamingConvention = '[A-Z][a-zA-Z_$0-9]*';
+	if ((language === 'Java' || language === 'Groovy') && !name.match(patternJavaNamingConvention)) {
+		return `The filename needs to follow the ${language} naming convention. I.e. ${patternJavaNamingConvention}`;
+	}
+	return undefined;
+}
+
+function getFileExtensionForLanguage(language: string): string {
+	const extension = LANGUAGES_WITH_FILENAME_EXTENSIONS.get(language);
+	if (extension) {
+		return extension;
+	} else {
+		return 'unknown';
+	}
+}
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -35,6 +35,7 @@ import { ChildProcess } from 'child_process';
 import { LogsPanel } from './logsWebview';
 import * as logUtils from './logUtils';
 import {checkKamelNeedsUpdate, version, handleChangeRuntimeConfiguration} from './versionUtils';
+import * as NewIntegrationFileCommand from './commands/NewIntegrationFileCommand';
 
 export const DELAY_RETRY_KUBECTL_CONNECTION = 1000;
 
@@ -156,6 +157,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<void> 
 				});
 			}
 		});
+
+		vscode.commands.registerCommand('camelk.integrations.createNewIntegrationFile', NewIntegrationFileCommand.create);
 
 	});
 
--- a/src/test/suite/commands/NewIntegrationFileCommand.test.ts
+++ b/src/test/suite/commands/NewIntegrationFileCommand.test.ts
@@ -1,0 +1,108 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+import { expect } from "chai";
+import * as fs from 'fs';
+import * as sinon from 'sinon';
+import * as vscode from 'vscode';
+import { fail } from "assert";
+
+const waitUntil = require('async-wait-until');
+
+suite('New Apache Camel K integration file', function() {
+
+	let showQuickpickStub: sinon.SinonStub;
+	let showInputBoxStub: sinon.SinonStub;
+	let showWorkspaceFolderPickStub: sinon.SinonStub;
+	let createdFile: vscode.Uri;
+
+	setup(() => {
+		showQuickpickStub = sinon.stub(vscode.window, 'showQuickPick');
+		showInputBoxStub = sinon.stub(vscode.window, 'showInputBox');
+		showWorkspaceFolderPickStub = sinon.stub(vscode.window, 'showWorkspaceFolderPick');
+	});
+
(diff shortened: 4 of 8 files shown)


apupier:
"please note that a kubeconfigf needs to be available currently," --> ahaha and of course it is not available on Travis...

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+import * as fs from 'fs';
+import * as path from 'path';
+import * as vscode from 'vscode';
+import * as kamel from './../kamel';
```
lhein, line 22: `./` seems obsolete. `../kamel` should be actually enough?

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+
+const validFilename = require('valid-filename');
+
+const LANGUAGES = ['Java', 'xml', 'Yaml', 'Groovy', 'JavaScript', 'Kotlin'];
```
lhein, line 26: Why not use Xml or XML instead of xml?

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export async function create() {
+	let language = await vscode.window.showQuickPick(LANGUAGES);
```
lhein, line 29: `language` is never changed again and hence can be a const instead.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+	if (language) {
+		let workspaceFolder = await vscode.window.showWorkspaceFolderPick();
```
lhein, line 31: `workspaceFolder` is never changed again and hence can be a const instead.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+		if (workspaceFolder) {
+			let filename = await vscode.window.showInputBox({
```
lhein, line 33: `filename` is never changed again and hence can be a const instead.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+				prompt: 'Please provide a name for the new file (without extension)',
+				validateInput: (name: string) => {
+					return validateFileName(name, language as string, workspaceFolder as vscode.WorkspaceFolder);
```
lhein, line 36: You can omit `as string` if you define the const using a datatype definition. for instance `const language : string = await ....`.

lhein: same applies to the folder

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+				}
+			});
+			if (filename) {
+				let kamelExe = kamel.create();
```
lhein, line 40: What is kamelExe? Is it the executable or is it a path to the executable? also kamelExe is never changed again, so it could be a const instead.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+				let fp: string = computeFullpath(language, workspaceFolder, filename);
```
lhein, line 41: what is fp? also fp is never changed again so it could be a const instead.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+				kamelExe.invoke(`init ${fp}`);
+			}
+		}
+	}
+}
+
+function computeFullpath(language: string, workspaceFolder: vscode.WorkspaceFolder, filename: string) {
+	let extension = getFileExtensionForLanguage(language);
```
lhein, line 49: `extension` is never changed again. can be a const.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+	return path.join(workspaceFolder.uri.fsPath, `${filename}.${extension}`);
```
lhein, line 50: path works fine with string folders too. Maybe its worth to change the parameter to string instead of WorksapceFolder? Where do u see the advantage in being that strict?

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export function validateFileName(name: string, language: string, workspaceFolder: vscode.WorkspaceFolder): string | undefined {
+	if (!name || name === '') {
```
lhein, line 54: I am not fully sure but I think `if (!name) will check for `undefined` and `empty` string. Maybe worth a check.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+		return 'Please provide a name for the new file (without extension)';
+	}
+	let fp: string = computeFullpath(language, workspaceFolder, name);
```
lhein, line 57: What is `fp`? Also it could be defined as a const because it never changes again.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+	if (fs.existsSync(fp)) {
+		return 'There is already a file with the same name. Please choose a different name.';
+	}
+	if (!validFilename(name)) {
+		return 'The filename is invalid.';
+	}
+	const patternJavaNamingConvention = '[A-Z][a-zA-Z_$0-9]*';
+	if ((language === 'Java' || language === 'Groovy') && !name.match(patternJavaNamingConvention)) {
+		return `the filename needs to follow the ${language} naming convention. I.e. ${patternJavaNamingConvention}`;
```
lhein, line 66: The sentence should start with a capital letter.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,87 @@
...
+	}
+	return undefined;
+}
+
+function getFileExtensionForLanguage(language: string): string {
```
lhein, line 71: How about using a map for storing the language to extension relation? It could make this function much shorter.

src/extension.ts
```diff
@@ -157,6 +158,10 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
 			}
 		});
 
+		vscode.commands.registerCommand('camelk.integrations.createNewIntegrationFile', async() => {
```
lhein, line 161: Optional: 
You can simply reference a function here instead of writing an async anonymous arrow function. 
Example: 
`context.subscriptions.push(vscode.commands.registerCommand('datavirt.create.vdb', createVDBCommand));`

src/test/suite/commands/NewIntegrationFileCommand.test.ts
```diff
@@ -0,0 +1,103 @@
...
+	});
+
+	test('Can create a new JavaScript integration file', async function() {
+		await testIntegrationFileCreation('TestCreation.js', 'JavaScript', 'TestCreation');
+	});
+
+	async function testIntegrationFileCreation(expectedFileNameWithExtension: string, languageToPick: string, providedFilename: string) {
+		expect(await vscode.workspace.findFiles(expectedFileNameWithExtension)).to.be.an('array').that.is.empty;
+		showQuickpickStub.onFirstCall().returns(languageToPick);
+		const workspaceFolder = (vscode.workspace.workspaceFolders as vscode.WorkspaceFolder[])[0];
```
lhein, line 75: using string instead of WorkspaceFolder can make life easier here.

src/commands/NewIntegrationFileCommand.ts
bfitzpat: It's in a deeper folder, so the path has to climb an additional level. I think it's ok

src/commands/NewIntegrationFileCommand.ts
apupier: XML sounds better effectively: https://fr.wikipedia.org/wiki/Extensible_Markup_Language

src/commands/NewIntegrationFileCommand.ts
apupier: Brian is right. It is required as it is in a deeper folder

src/commands/NewIntegrationFileCommand.ts
apupier: it is the interface provided by this extension to the Kamel executable

src/commands/NewIntegrationFileCommand.ts
lhein: `./` is the current folder
`./../` is a complicated way to write `../`

src/commands/NewIntegrationFileCommand.ts
lhein: I find this name is a bit misleading.

apupier:
> Will look more detailed tomorrow.

no need to rush on this one. other work is required before I'm working back on it.

src/commands/NewIntegrationFileCommand.ts
apupier: my bad, I read too fast the feedback.
Will need to look if there is adifference or a recommended way in TypeScript but effectively `../` is easier to read

lhein: It is really no black magic. Its just the way it works on any operating system. Go to a shell and type both and you will see u end up in the same folder.

src/commands/NewIntegrationFileCommand.ts
apupier: ok.
For this PR, i think we need to keep it as-is to stay coherent with other parts of the codebase https://github.com/camel-tooling/vscode-camelk/search?q=kamelExe&unscoped_q=kamelExe
I created https://issues.redhat.com/browse/FUSETOOLS2-348

src/commands/NewIntegrationFileCommand.ts
apupier: it is not that straightforward. In examples on Typescript language, they are using `./` notation for same level files https://www.typescriptlang.org/docs/handbook/modules.html
I can understand that to stay coherent with that, a good practice would be to keep it.

I don't have opinion on that. Would need to find reference.

lhein: Yeah, I see your point. Now I am interested too.

lhein: Check out https://www.typescriptlang.org/docs/handbook/module-resolution.html

lhein: image

apupier: cool, that's a good reference

src/commands/NewIntegrationFileCommand.ts
apupier: We are retrieving a WorkspaceFolder from `vscode.window.showWorkspaceFolderPick()`
it factorizes the `workspaceFolder.uri.fsPath`, otherwise we would have to call it before calling the `computeFullpath` method.

src/test/suite/commands/NewIntegrationFileCommand.test.ts
apupier: we need a WorkspaceFolder to mock the VS Code API

apupier:
this other PR rebased on branch which provides Minikube is showing that tests can be executed fine: https://github.com/camel-tooling/vscode-camelk/pull/377

README.md
```diff
@@ -106,7 +115,7 @@ To update the state of your currently deployed integrations, hover over the **Ap
 
 ![Apache Camel K Integrations view - Refresh](images/camelk-integrations-view-refresh-action.jpg)
 
-## Creating a new Camel K Integration with multiple parameters
+## Creating a new Camel K Integration task configuration with multiple parameters
```
apupier, line 118: I changed it in this PR as I was finding that it was ambiguous with the "new Camel K Integration file" added feature

src/commands/NewIntegrationFileCommand.ts
apupier: this language has such strange behavior

lhein:
When the workspace folder picker is displayed and I accedently type something then the result list is empty and there is no explanation what was this dialog about. It would be nice if you could provide a prompt for that dialog. Same applies to the language selection dialog.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,85 @@
...
+			if (filename) {
+				let kamelExe = kamel.create();
+				let newFileFullPath: string = computeFullpath(language, workspaceFolder, filename);
+				kamelExe.invoke(`init ${newFileFullPath}`);
+			}
+		}
+	}
+}
+
+function computeFullpath(language: string, workspaceFolder: vscode.WorkspaceFolder, filename: string) {
```
lhein, line 55: How about defining the return type `string` for the function?

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,85 @@
...
+const LANGUAGES_WITH_FILENAME_EXTENSIONS = new Map([
+	['Java', 'java'],
+	['XML', 'xml'],
+	['Yaml', 'yaml'],
+	['Groovy', 'groovy'],
+	['JavaScript', 'js'],
+	['Kotlin', 'kts']]);
+const LANGUAGES = Array.from(LANGUAGES_WITH_FILENAME_EXTENSIONS.keys());
+
+export async function create() {
```
lhein, line 35: How about defining the return type `Promise<void>` for this function?

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,85 @@
...
+		const workspaceFolder = await vscode.window.showWorkspaceFolderPick();
+		if (workspaceFolder) {
+			const filename = await vscode.window.showInputBox({
+				prompt: 'Please provide a name for the new file (without extension)',
+				validateInput: (name: string) => {
+					return validateFileName(name, language, workspaceFolder);
+				}
+			});
+			if (filename) {
+				let kamelExe = kamel.create();
```
lhein, line 47: kamelExe is not changed and could be made a const.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -0,0 +1,85 @@
...
+				let newFileFullPath: string = computeFullpath(language, workspaceFolder, filename);
```
lhein, line 48: newFileFullPath is never changed and could be made a const.

src/test/suite/commands/NewIntegrationFileCommand.test.ts
```diff
@@ -0,0 +1,108 @@
...
+	test('Can create a new Kotlin integration file', async function() {
+		await testIntegrationFileCreation('TestCreation.kts', 'Kotlin', 'TestCreation');
+	});
+
+	test('Can create a new JavaScript integration file', async function() {
+		await testIntegrationFileCreation('TestCreation.js', 'JavaScript', 'TestCreation');
+	});
+
+	async function testIntegrationFileCreation(expectedFileNameWithExtension: string, languageToPick: string, providedFilename: string) {
+		expect(await vscode.workspace.findFiles(expectedFileNameWithExtension)).to.be.an('array').that.is.empty;
```
lhein, line 74: TSLint shows a warning on that line.
`unused expression, expected an assignment or function call`

apupier: this is a false-positive as far I understand it. The expect API is working like that and tslint is nto able to understand it

apupier:
> After creating the new file it would be nice if it would open up in an editor so I can start coding. What do you think?

agree but unfortunately it requires a lot of changes: issue created to handled it https://issues.redhat.com/browse/FUSETOOLS2-303 as mentioned in the first comment of the PR

src/commands/NewIntegrationFileCommand.ts
apupier: I really hate this refactor support in VS Code which is not adding the full signature

lhein:
> > After creating the new file it would be nice if it would open up in an editor so I can start coding. What do you think?
> 
> agree but unfortunately it requires a lot of changes: issue created to handled it https://issues.redhat.com/browse/FUSETOOLS2-303 as mentioned in the first comment of the PR

Ah, I missed that. Great.

lhein:
> When the workspace folder picker is displayed and I accidently type something then the result list is empty and there is no explanation what this dialog was about. It would be nice if you could provide a prompt for that dialog. Same applies to the language selection dialog.

I still think its worth to add prompts to all input dialogs. It will improve the user experience.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -33,9 +33,10 @@ const LANGUAGES_WITH_FILENAME_EXTENSIONS = new Map([
 const LANGUAGES = Array.from(LANGUAGES_WITH_FILENAME_EXTENSIONS.keys());
 
 export async function create() : Promise<void> {
-	const language = await vscode.window.showQuickPick(LANGUAGES);
+	const language = await vscode.window.showQuickPick(LANGUAGES, {placeHolder:'Please select the language in which the new file will be generated.'});
 	if (language) {
-		const workspaceFolder = await vscode.window.showWorkspaceFolderPick();
+		const workspaceFolder = await vscode.window.showWorkspaceFolderPick(
+			{placeHolder: 'Please select the Workspace folder in which the new file will be created.'});
```
lhein, line 39: I think workspace should not start with capital letter. Would you agree?

apupier:
> When the workspace folder picker is displayed and I accedently type something then the result list is empty

unfortunately, this sounds a normal behavior for list to select in VS Code. The field is acting as a filter.

> there is no explanation what this dialog was about. It would be nice if you could provide a prompt for that dialog. Same applies to the language selection dialog.

placeholder provided

src/commands/NewIntegrationFileCommand.ts
apupier: in the UI and documentation there is a mix of capital letters and not capital letters.
At least needs to be either "Workspace Folder" or "workspace folder". I changed to "workspace folder"

apupier:
thanks for the great review

PR #358, opened 2020-03-24
FUSETOOLS2-305 use cache for latest Camel version available

apupier:
- the cache needs a restart of VS Code to be updated
- given that the version is modified no more than once a month, I don't
think that it is a problem
- the idea is also to avoid 403 rate limit exceeded when playing tests
on CIs

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -24,4 +24,4 @@
 
 # Note about test execution and GitHub API access
 
-Access to GitHub API has rate limit. This rate limit is lower for unauthenticated request. On CI, this rate limit was often hit (either Travis or Jenkins). To avoid that, a GitHub token needs to be provided through VSCODE_CAMELK_GITHUB_TOKEN environment variable. This token doesn't require any specific rights.
+Access to GitHub API has a rate limit. This rate limit is lower for unauthenticated requests. On CI, this rate limit was often hit (either Travis or Jenkins). To avoid that, a GitHub token needs to be provided through VSCODE_CAMELK_GITHUB_TOKEN environment variable. This token doesn't require any specific rights.
--- a/Contributing.md
+++ b/Contributing.md
@@ -21,3 +21,7 @@
   * Run 'npm install' so that the package-lock.json is updated
   * Push changes in a PR
   * Follow PR until it is approved/merged
+
+# Note about test execution and GitHub API access
+
+Access to GitHub API has rate limit. This rate limit is lower for unauthenticated request. On CI, this rate limit was often hit (either Travis or Jenkins). To avoid that, a GitHub token needs to be provided through VSCODE_CAMELK_GITHUB_TOKEN environment variable. This token doesn't require any specific rights.
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -128,15 +128,18 @@ function updateStatusBarItem(sbItem : vscode.StatusBarItem, text: string, toolti
 }
 
 export async function installKamel(context: vscode.ExtensionContext): Promise<Errorable<null>> {
-	const latestversion = await versionUtils.getLatestCamelKVersion();
-	if (failed(latestversion)) {
-		return { succeeded: false, error: latestversion.error };
-	}
-
-	let versionToUse = latestversion.result.trim();
+	let versionToUse: string;
 	let runtimeVersionSetting = vscode.workspace.getConfiguration().get(config.RUNTIME_VERSION_KEY) as string;
 	if (runtimeVersionSetting && runtimeVersionSetting.toLowerCase() !== versionUtils.version.toLowerCase()) {
 		versionToUse = runtimeVersionSetting;
+	} else {
+		const latestversion = await versionUtils.getLatestCamelKVersion();
+		if (failed(latestversion)) {
+			extension.shareMessageInMainOutputChannel(`Cannot retrieve latest available Camel version and none specified in settings. Will fallback to use the default ${versionUtils.version}`);
+			versionToUse = versionUtils.version;
+		} else {
+			versionToUse = latestversion.result.trim();
+		}
 	}
 
 	await versionUtils.checkKamelNeedsUpdate(versionToUse).then((needsUpdate) => {
--- a/src/installer.ts
+++ b/src/installer.ts
@@ -129,13 +129,13 @@ function updateStatusBarItem(sbItem : vscode.StatusBarItem, text: string, toolti
 
 export async function installKamel(context: vscode.ExtensionContext): Promise<Errorable<null>> {
 	let versionToUse: string;
-	let runtimeVersionSetting = vscode.workspace.getConfiguration().get(config.RUNTIME_VERSION_KEY) as string;
+	const runtimeVersionSetting = vscode.workspace.getConfiguration().get(config.RUNTIME_VERSION_KEY) as string;
 	if (runtimeVersionSetting && runtimeVersionSetting.toLowerCase() !== versionUtils.version.toLowerCase()) {
 		versionToUse = runtimeVersionSetting;
 	} else {
 		const latestversion = await versionUtils.getLatestCamelKVersion();
 		if (failed(latestversion)) {
-			extension.shareMessageInMainOutputChannel(`Cannot retrieve latest available Camel version and none specified in settings. Will fallback to use the default ${versionUtils.version}`);
+			extension.shareMessageInMainOutputChannel(`Cannot retrieve latest available Camel version and none has been specified in settings. Will fall back to use the default ${versionUtils.version}`);
 			versionToUse = versionUtils.version;
 		} else {
 			versionToUse = latestversion.result.trim();
--- a/src/test/suite/install.test.ts
+++ b/src/test/suite/install.test.ts
@@ -50,7 +50,11 @@ suite("ensure install methods are functioning as expected", function() {
 		installKubectlSpy.resetHistory();	
 	});
 
-	test("ensure cli version checking works correctly", async function() {
+	var testVar = test("ensure cli version checking works correctly", async function() {
+		if(process.env.VSCODE_CAMELK_GITHUB_TOKEN === undefined) {
+			testVar.skip();
+		}
+
 		await versionUtils.checkKamelNeedsUpdate('bogusversion').then( () => {
 			assert.fail('cli version checking, negative case worked');
 		}).catch( (error) => {
--- a/src/test/suite/install.test.ts
+++ b/src/test/suite/install.test.ts
@@ -51,7 +51,7 @@ suite("ensure install methods are functioning as expected", function() {
 	});
 
 	var testVar = test("ensure cli version checking works correctly", async function() {
-		if(process.env.VSCODE_CAMELK_GITHUB_TOKEN === undefined) {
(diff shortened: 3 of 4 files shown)


apupier:
the url was called 3 times during test execution. With cache version, it is now called a single time.

apupier:
even with a single call, this is still not enough:
`AssertionError [ERR_ASSERTION]: Failed to establish Apache Camel K stable version: Response code 403 (Forbidden)`

seems like between 2 builds, we need to have a longer delay...

apupier:
rate limiting on api.github with unauthenticated si limited at 60 requests per hour, it is by IP. Not sure how request from travis are coming from same:different IP. (same job?build?project?organization?random?)

https://developer.github.com/v3/#rate-limiting
`For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the user making requests.`

apupier:
we could use conditional requests to see if there was a modification since "last time", "last time" would be a version that we push on the repository.

https://developer.github.com/v3/#conditional-requests

apupier:
given that it is still failing when using conditional request, I suppose that conditional request doesn't count in rate limit only for authenticated connection. I cannot find clear statement on that in github documentation

apupier:
I (try to) configured to use authenticated method using a Token.
Unfortunately, the token cannot be passed on branches outside of vscode-camelk repository...
```
Encrypted environment variables have been removed for security reasons.
See https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions
```

it means that for pull requests, we will still have the same issue!

apupier:
we are not seeing anymore the failure but it is just that there is the same sha1 for the other branch that i created on camel-tooling organization.

When using authenticated we can notice that it is working (I relaunched the build several times).

So one possible way is to disable test or mock call when token is not available. This way, the test will be still played with real connection on master. We will detect potential connection problem a bit late but not never, which is far better.

apupier:
I disabled the test.
There were other test failing because, there was no fallback in case the api github wasn't reachable. I modified to fallback to default version; Also, avoid to call the api github if a specific runtime version is specified in settings as it is preponderant.

src/versionUtils.ts
```diff
@@ -18,30 +18,34 @@
...
 import * as kamelCli from './kamel';
-import * as fs from 'fs';
-import {platformString, kamelUnavailableRejection} from './installer';
+import { platformString, kamelUnavailableRejection } from './installer';
 import fetch from 'cross-fetch';
 
-export const version : string = '1.0.0-RC2'; //need to retrieve this if possible, but have a default
+export const version: string = '1.0.0-RC2'; //need to retrieve this if possible, but have a default
+/*
+* Can be retrieved using `curl -i https://api.github.com/repos/apache/camel-k/releases/latest` and searchign for "last-modified" attribute
```
apupier: argh typo, will update it

Contributing.md
```diff
@@ -21,3 +21,7 @@
   * Run 'npm install' so that the package-lock.json is updated
   * Push changes in a PR
   * Follow PR until it is approved/merged
+
+# Note about test execution and GitHub API access
+
+Access to GitHub API has rate limit. This rate limit is lower for unauthenticated request. On CI, this rate limit was often hit (either Travis or Jenkins). To avoid that, a GitHub token needs to be provided through VSCODE_CAMELK_GITHUB_TOKEN environment variable. This token doesn't require any specific rights.
```
lhein: either "has a rate limit" or "has rate limits"

lhein: "is lower for unauthenticated requests"

src/installer.ts
```diff
@@ -128,15 +128,18 @@ function updateStatusBarItem(sbItem : vscode.StatusBarItem, text: string, toolti
...
 
 export async function installKamel(context: vscode.ExtensionContext): Promise<Errorable<null>> {
-	const latestversion = await versionUtils.getLatestCamelKVersion();
-	if (failed(latestversion)) {
-		return { succeeded: false, error: latestversion.error };
-	}
-
-	let versionToUse = latestversion.result.trim();
+	let versionToUse: string;
 	let runtimeVersionSetting = vscode.workspace.getConfiguration().get(config.RUNTIME_VERSION_KEY) as string;
```
lhein: if runtimeVersionSetting is never changed again then make it a const instead of a let

lhein: I know its not part of your changes but I wanted to mention it anyways.

src/test/suite/install.test.ts
```diff
@@ -50,7 +50,11 @@ suite("ensure install methods are functioning as expected", function() {
 		installKubectlSpy.resetHistory();	
 	});
 
-	test("ensure cli version checking works correctly", async function() {
+	var testVar = test("ensure cli version checking works correctly", async function() {
+		if(process.env.VSCODE_CAMELK_GITHUB_TOKEN === undefined) {
```
lhein: could be simplified by 
		if(!process.env.VSCODE_CAMELK_GITHUB_TOKEN) {

src/versionUtils.ts
```diff
@@ -18,30 +18,34 @@
...
+* To be updated when updating the default "version" attribute
+*/
+const LAST_MODIFIED_DATE_OF_DEFAULT_VERSION: string = 'Fri, 28 Feb 2020 07:24:17 GMT';
+let latestVersionFromOnline: string;
 
-async function testVersionAvailable(versionToUse : string): Promise<boolean> {
-	return new Promise<boolean>( async (resolve) => {
+async function testVersionAvailable(versionToUse: string): Promise<boolean> {
```
lhein: if you return a Promise this function is automatically async. There is no real need to specify it twice. Same applies you you don't specify a return value, then async will make it a Promise automatically.

src/versionUtils.ts
```diff
@@ -101,31 +105,46 @@ export async function pingGithubUrl(urlStr: string) : Promise<boolean> {
...
-	let latestJSON = JSON.parse(rawtext);
-	let tagName = latestJSON.tag_name;
-	if (tagName) {
-		return { succeeded: true, result: tagName };
+	if (latestVersionFromOnline) {
+		return { succeeded: true, result: latestVersionFromOnline };
+	} else {
+		const latestURL = 'https://api.github.com/repos/apache/camel-k/releases/latest';
+		const headers = [['If-Modified-Since', LAST_MODIFIED_DATE_OF_DEFAULT_VERSION]];
+		let githubToken = process.env.VSCODE_CAMELK_GITHUB_TOKEN;
```
lhein: use const instead of let

src/versionUtils.ts
```diff
@@ -101,31 +105,46 @@ export async function pingGithubUrl(urlStr: string) : Promise<boolean> {
...
+		if(githubToken) {
+			headers.push(['Authorization', `token ${githubToken}`]);
+		}
+		const res = await fetch(latestURL, { headers: headers });
+		if (res.status === 200) {
+			let latestJSON = await res.json();
```
lhein: if not changed afterwards use const

src/versionUtils.ts
```diff
@@ -101,31 +105,46 @@ export async function pingGithubUrl(urlStr: string) : Promise<boolean> {
...
+			let tagName = latestJSON.tag_name;
```
lhein: if not changed afterwards use const

src/versionUtils.ts
```diff
@@ -101,31 +105,46 @@ export async function pingGithubUrl(urlStr: string) : Promise<boolean> {
...
+		}
 	}
-	return { succeeded: false, error: [`Failed to retrieve latest Apache Camel K version tag from : ${latestURL}`] };
 }
 
-function checkKamelCLIVersion() : Promise<string> {
-	return new Promise<string>( async (resolve, reject) => {
+function checkKamelCLIVersion(): Promise<string> {
+	return new Promise<string>(async (resolve, reject) => {
 		let kamelLocal = kamelCli.create();
```
lhein: const instead of let

src/versionUtils.ts
apupier: it is not part of the modification that I made.
the change doesn't seem trivial for me so I prefer tol keep it as-is in this PR

src/installer.ts
apupier: updated as it is a trivial change

src/versionUtils.ts
apupier: updated as a it is a trivial change even if it wasn't part of modified code

src/test/suite/install.test.ts
apupier: it doesn't seem to be the same. The test is no more skipped: https://travis-ci.org/github/camel-tooling/vscode-camelk/builds/666721723#L637

lhein: which would mean that the env variable is somehow set

apupier: effectively set in this build: https://travis-ci.org/github/camel-tooling/vscode-camelk/builds/666721723#L187 don't understand why it is set as it is not on the cameltooling repository

apupier: ok, there is a pending test when the env variable is not set: https://travis-ci.org/github/camel-tooling/vscode-camelk/builds/666756002#L639

lhein: that seems to be the correct behavior then. Anyways I am a bit concerned about the other exceptions in the test output....

`rejected promise not handled within 1 second: TypeError: Cannot read property 'catch' of undefined`

lhein: that looks like we have somewhere an undefined value and try to access member vars which then causes an uncatched error which makes the promise time out

lhein: seems to be in integrationExplorer.test.ts

apupier: Unless you are pointing to a new one, this was not introduced by this PR.
see https://travis-ci.org/github/camel-tooling/vscode-camelk/builds/664766935#L546 for a maste rbuild with this error

src/installer.ts
```diff
@@ -128,15 +128,18 @@ function updateStatusBarItem(sbItem : vscode.StatusBarItem, text: string, toolti
...
-	let versionToUse = latestversion.result.trim();
-	let runtimeVersionSetting = vscode.workspace.getConfiguration().get(config.RUNTIME_VERSION_KEY) as string;
+	let versionToUse: string;
+	const runtimeVersionSetting = vscode.workspace.getConfiguration().get(config.RUNTIME_VERSION_KEY) as string;
 	if (runtimeVersionSetting && runtimeVersionSetting.toLowerCase() !== versionUtils.version.toLowerCase()) {
 		versionToUse = runtimeVersionSetting;
+	} else {
+		const latestversion = await versionUtils.getLatestCamelKVersion();
+		if (failed(latestversion)) {
+			extension.shareMessageInMainOutputChannel(`Cannot retrieve latest available Camel version and none specified in settings. Will fallback to use the default ${versionUtils.version}`);
```
bfitzpat: none specified in settings" should be "none has been specified in settings. Will fall back" (fall back is two words and a bit clearer in the first sentence)

src/test/suite/install.test.ts
lhein: ok

apupier:
Thanks for all the feedback

PR #360, opened 2020-03-25
Fusetools2 305 use cache to retrievelatest version

apupier:
same branch but from main repo to have environment variables available


apupier:
closing, was used just to show differences when on camel-toolign organization o rnot

PR #385, opened 2020-04-22
Didact ck solx

bfitzpat:
This PR does a couple of things:

1. It updates two of the main Camel K commands (start integration and new integration file) so that we can leverage them in Didact. Essentially this means adapting those command functions to handle an arbitrary number of arguments. In the standard, in-tooling case, usually the first argument is a vscode.Uri for the resource and then any additional arguments are passed beyond that. In the Didact case, we wanted a way to pass in default values. 

2. It adds a didact tutorial for Camel K (if Didact is installed) walking through creating an integration, starting an integration, and looking at the logs for the integration. Didact is NOT required, but if it is there, the new tutorial shows up in an "Apache Camel K" folder in the Tutorials view.


bfitzpat:
@lhein This should give you some ideas for how to implement a similar DV didact tutorial when you're ready

didact/solx-camel-k-basic/solx-camel-k-basic-alt.didact.adoc
```diff
@@ -76,35 +93,34 @@ While our integration is running in `Dev Mode`, we can modify it and see those c
...
-* Hover over the integration name to see its current state in the tooltip. 
-* Right-click on the running integration to `Remove Apache Camel K Integration` and undeploy it.
-* Right-click on the running integration to `Follow log for running Apache Camel K Integration`
-* Right-click on the running integration to `Follow kit builder log for running Apache Camel K integration`
+* Hover over the `Apache Camel K Integrations` view and click the `Open Apache Camel K Operator Log` button to view it. (link:didact://?commandId=camelk.integrations.openOperatorLog[Execute^])
+* Hover over the integration name to see its current state in the tooltip. (link:didact://?commandId=camelk.integrations.selectFirstNode[Execute to select first integration in tree^])
+* Right-click on the running integration to `Remove Apache Camel K Integration` and undeploy it. (link:didact://?commandId=camelk.integrations.remove[DANGER! WILL REMOVE RUNNING INTEGRATION])
+* Right-click on the running integration to `Follow kit builder log for running Apache Camel K integration` (link:didact://?commandId=camelk.integrations.kitlog[Execute to view log for first integration in tree^])
```
djelinek, line 111: one missing capital letter "Follow kit builder log for running Apache Camel K Integration"

didact/solx-camel-k-basic/solx-camel-k-basic-alt.didact.adoc
```diff
@@ -76,35 +93,34 @@ While our integration is running in `Dev Mode`, we can modify it and see those c
...
 
-While we are running in `Dev Mode`, all our logged output goes to the main `Apache Camel K` Output channel, but if the integration is running in another mode (such as `Basic`), we can explicitly open a new Output channel to follow the log for that running integration.
+While we are running in `Dev Mode`, all our logged output goes to the main `Apache Camel K` Output channel, but if the integration is running in another mode (such as `Basic`), we can explicitly open a new log window to see events for that running integration.
 
-* To open a separate log window for your route, right-click on the deployed integration 'hello-world' in the `Apache Camel K Integrations` view and select `Follow log for running Apache Camel K integration`. 
+* To open a separate log window for your route, right-click on the deployed integration 'hello-world' in the `Apache Camel K Integrations` view and select `Follow log for running Apache Camel K integration`. (link:didact://?commandId=camelk.integrations.log[Execute to view log for first integration in tree^])
```
djelinek, line 115: one missing capital letter "Follow log for running Apache Camel K Integration"

didact/solx-camel-k-basic/solx-camel-k-basic-alt.didact.adoc
```diff
@@ -76,35 +93,34 @@ While our integration is running in `Dev Mode`, we can modify it and see those c
...
 
 You should see your Camel route injecting a new message in the log every few seconds.
 
 == Did it not work?
 
-Remove the simple directory and try the steps again. If it's still not working:
+Remove the integration and try the steps again. If it's still not working:
 
 . Validate that you have access to the Apache Camel K CLI (via terminal or via command)
 . Check if something went wrong when starting the integration. View the Apache Camel K Operator log from the view toolbar or Kit Builder log avaialble when you right-click the integration
```
djelinek, line 126: typo: "avaialble" -> "available"

bfitzpat:
Thanks @djelinek !

djelinek:
You're welcome. yeah I saw the DRAFT label but I dived into reading and spot few trivials :-)

bfitzpat:
Added a test for passing arguments in the Didact way, but since starting the integration is an in itself an integration test, I don't have a good way to do that in a unit test at this time. Definitely something to handle in a UI test perhaps?

bfitzpat:
Will spin the solution pattern work into its own PR and make some changes to the "first integration" example, but this is almost in a good state.

src/IntegrationUtils.ts
```diff
@@ -59,13 +60,55 @@ const choiceList = [
 
  let childProcessMap : Map<string, child_process.ChildProcess>;
 
- export function startIntegration(context: vscode.Uri): Promise<boolean> {
+ export function startIntegration(...args: any[]): Promise<boolean> {
```
lhein, line 63: I am wondering if we should not better do `startIntegration(context: vscode.Uri, ...args: any[])` instead to keep the didact in a separate flow.

bfitzpat: I can definitely try that.

src/IntegrationUtils.ts
```diff
@@ -59,13 +60,55 @@ const choiceList = [
...
 	return new Promise <boolean> ( async (resolve, reject) => {
-		const choice : string | undefined = await vscode.window.showQuickPick(choiceList, {
-			placeHolder: 'Select the type of Apache Camel K Integration'
-		});
+		let context : vscode.Uri | undefined = undefined;
+		let inChoice : string | undefined = undefined;
+
+		// with no arguments, try working with the selected file
+		if (!args || args.length === 0) {
+			await getCurrentFileSelectionPath().then((fileUri) => {
```
lhein, line 70: I wouldn't mix await and then / catch. 
`try {
   context = await getCurrentFileSelectionPath();
   inChoice = devModeIntegration;
} catch (error) {
   reject('No arguments provided to start integration function call');
   return;	
}

src/IntegrationUtils.ts
```diff
@@ -509,4 +552,35 @@ export function killChildProcessForIntegration(integration:string) : Promise<boo
...
+			return basicIntegration;
+		}
+	}
+	// other choices not supported at present because they will require additional inputs
+	return undefined;
+ }
+
+ export async function getCurrentFileSelectionPath(): Promise<vscode.Uri> {
+	if (vscode.window.activeTextEditor)
+	{
```
lhein, line 572: should be better after the `)` ?

bfitzpat: formatting-wise? maybe

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -32,27 +32,50 @@ const LANGUAGES_WITH_FILENAME_EXTENSIONS = new Map([
...
-					return validateFileName(name, language, workspaceFolder);
-				}
-			});
-			if (filename) {
-				const kamelExe = kamel.create();
-				const newFileFullPath: string = computeFullpath(language, workspaceFolder, filename);
-				await kamelExe.invoke(`init "${newFileFullPath}"`);
-				const textDocument = await vscode.workspace.openTextDocument(newFileFullPath);
-				await vscode.window.showTextDocument(textDocument);
+export async function create(...args: any[]) : Promise<void> {
```
lhein, line 35: same as with the other command. I would separate the didact ones from the usual expected one.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -32,27 +32,50 @@ const LANGUAGES_WITH_FILENAME_EXTENSIONS = new Map([
...
+
+	let language : string | undefined;
+	let workspaceFolder : vscode.WorkspaceFolder | undefined = undefined;
```
lhein, line 38: I think that `undefined` is the default anyways when not assigning a value so it could be removed.

src/extension.ts
```diff
@@ -188,18 +217,18 @@ export function hideStatusLine(): void {
 }
 
 // start the integration file
-async function runTheFile(context: vscode.Uri) {
-	await startIntegration(context)
+async function runTheFile(...args: any[]) {
+	await startIntegration(args)
```
lhein, line 221: better not mix await with then/catch

src/extension.ts
```diff
@@ -188,18 +217,18 @@ export function hideStatusLine(): void {
...
 }
 
 // start an integration from a file
-function startIntegration(context: vscode.Uri): Promise<any> {
+function startIntegration(...args: any[]): Promise<any> {
 	return new Promise <any> ( async (resolve, reject) => {
 		setStatusLineMessageAndShow(`Starting new Apache Camel K Integration...`);
 		utils.shareMessage(mainOutputChannel, "Starting new integration via Kamel executable.");
-		await integrationutils.startIntegration(context)
+		await integrationutils.startIntegration(args)
```
lhein, line 231: better not mix await with then/catch

src/commands/NewIntegrationFileCommand.ts
bfitzpat: I can try it but I seem to remember running into some weirdness when I didn't explicitly specify `string | undefined`

lhein: I meant changing it to `let workspaceFolder : vscode.WorkspaceFolder | undefined;`

src/IntegrationUtils.ts
bfitzpat: So, regardless of what I put in the function argument declaration, the first argument always comes through as an array of arguments. It appears that execute isn't distinguishing between a context-only argument list and a mix of arguments... it's always passed an array of any[]...

`executeCommand<T>(command: string, ...rest: any[]): Thenable<T | undefined>`

We're getting the "rest" passed to the command function triggered when the command Id is called.

src/IntegrationUtils.ts
bfitzpat: Isn't the whole idea of resolve/reject that you can use then/catch?

src/extension.ts
bfitzpat: Bad things happen in here trying to switch to try/catch.

src/IntegrationUtils.ts
bfitzpat: This one I was able to switch to a try/catch.

src/extension.ts
bfitzpat: Again, bad things happen when I try to move this to try/catch. I think that the first case worked because it wasn't set up with a resolve/reject. This particular tangled combination in extension.ts is tied to that resolve/reject pattern.

bfitzpat:
@lhein I'm curious about your feedback regarding the `try/catch` approach vs. `function().then().catch()`. I thought the whole idea behind resolve/reject was that we could fall into this then/catch idea. And in places where I'm using resolve/reject it seems to work really well. So why would I change it?

src/commands/NewIntegrationFileCommand.ts
bfitzpat: see earlier comment about executeCommand

bfitzpat:
@lhein I moved the runthefile function to use a try/catch with the two await statements sequentially called. That seems to work fine in tests and local testing. Thanks for the offline clarification.

src/IntegrationUtils.ts
lhein: ah, my fault. Now I see that this is called directly by the command.

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,132 @@
...
+
+[Check out the Apache Camel K project documentation for more details about the framework.](https://camel.apache.org/camel-k/latest/index.html)
+
+## Apache Camel K in VS Code - Your First Integration
+
+What follows is a simple step-by-step process that helps you create and deploy a Camel K integration on a Minikube instance running locally.
+
+We will:
+
+* Create a folder with a sample Apache Camel integration written in Groovy
```
apupier, line 17: we should not provide a Groovy example as it not in the languages in priority. A java or xml would be better.

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,132 @@
...
+| [VS Code Extension Pack for Apache Camel by Red Hat is installed](didact://?commandId=vscode.didact.extensionRequirementCheck&text=extension-requirement-status$$redhat.apache-camel-extension-pack "Checks the VS Code workspace to make sure the extension pack is installed"){.didact} | *Status: unknown*{#extension-requirement-status} 	| [Click here to install](vscode:extension/redhat.apache-camel-extension-pack "Opens the extension page and provides an install link") |
+| [VS Code Tooling for Apache Camel K by Red Hat is installed](didact://?commandId=vscode.didact.extensionRequirementCheck&text=camelk-extension-requirement-status$$redhat.vscode-camelk "Checks the VS Code workspace to make sure the extension pack is installed"){.didact} | *Status: unknown*{#camelk-extension-requirement-status} 	| [Click here to install](vscode:extension/redhat.vscode-camelk "Opens the extension page and provides an install link") |
```
apupier, line 32: the VS Code Apache Camel K check can be removed it is coming with the Extension pack

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Your First Camel K Integration
+
+You can write an integration in one of several languages supported ([Groovy, Kotlin, JavaScript, Java, XML, YAML, etc.](https://camel.apache.org/camel-k/latest/languages/languages.html)), but today we're going to focus on Groovy which is one of the most common: Java.
```
apupier, line 36: `but today we're going to focus on Groovy which is one of the most common: Java.`

Groovy/Java ?

didact/camelk/first-integration.md
bfitzpat: As of today it is not listed as part of the extension pack - https://marketplace.visualstudio.com/items?itemName=redhat.apache-camel-extension-pack

src/IntegrationUtils.ts
```diff
@@ -509,4 +550,34 @@ export function killChildProcessForIntegration(integration:string) : Promise<boo
...
+	}
+	// other choices not supported at present because they will require additional inputs
+	return undefined;
+ }
+
+ export async function getCurrentFileSelectionPath(): Promise<vscode.Uri> {
+	if (vscode.window.activeTextEditor) {
+	  return vscode.window.activeTextEditor.document.uri;
+	}
+	else {
```
apupier, line 572: else on a newline

src/IntegrationUtils.ts
```diff
@@ -509,4 +550,34 @@ export function killChildProcessForIntegration(integration:string) : Promise<boo
...
+			return basicIntegration;
+		}
+	}
+	// other choices not supported at present because they will require additional inputs
+	return undefined;
+ }
+
+ export async function getCurrentFileSelectionPath(): Promise<vscode.Uri> {
+	if (vscode.window.activeTextEditor) {
+	  return vscode.window.activeTextEditor.document.uri;
```
apupier, line 570: mix of space and tabs

didact/camelk/first-integration.md
bfitzpat: Fixed

didact/camelk/first-integration.md
bfitzpat: Fixed

didact/camelk/first-integration.md
apupier: it is listed even though it is missing a specifc dot.

Screenshot from 2020-06-03 16-16-46

we can check that it is available here https://github.com/camel-tooling/vscode-camel-extension-pack/blob/350b0e126e3f0d1e75b50c6af6079937a29fbfe3/package.json#L28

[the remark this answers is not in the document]
bfitzpat: Open to suggestions on how to require this flexible dependency, though in principle I do agree. Not sure how to do that beyond add a manual step with a link to the installation directions. https://camel.apache.org/camel-k/latest/installation/installation.html

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,132 @@
...
+
+You can download the file yourself from the GitHub repo and then copy it into a folder in your workspace. Or you can create it with the tooling.
+
+Inside VS Code, press `F1` or `Ctrl+Shift+P` to bring up the Command Palette, and type `Create a new Apache Camel K Integration file`. When the command is selected, click Enter.
+
+1. Choose the language to use (choose `Java`).
+2. Choose the workspace folder (press Enter to select the current workspace root folder)
+3. Provide a name for the new file (be sure not to include the file extension or it will be repeated) (type `simple`)
+
+When you complete step 3, you should see the file `simple.java` appear in your workspace folder. [(Execute^)](didact://?commandId=camelk.integrations.createNewIntegrationFile&text=simple$$Java)
```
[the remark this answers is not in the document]
bfitzpat: Yes, which is a visual hint that other Didact writers have been using to indicate an action beyond the scope of Didact. Unless we come up with another approach, I'm going to leave the caret character in those Execute links.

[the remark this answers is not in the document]
bfitzpat: I will add it more consistently throughout the Didact file

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,132 @@
...
+
+```
+
+</details>
+
+## Step 2: Exploring the Groovy integration file
+
+Now that you have an integration file, let's take a quick look at it. If you created the file yourself, go ahead and open it now. Go to the Explorer activity (Ctrl+Shift+E) and look at the workspace folders listed.
+
+If you clicked the link to create integration file earlier, look for `simple.java` in the Explorer view.
```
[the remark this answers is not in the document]
bfitzpat: Yes and no. There's a possibility they created it by hand, so they still need to make sure they open the right file in the Explorer. I'll massage that text a bit.

didact/camelk/first-integration.md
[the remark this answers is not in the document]
apupier: I think we can provide a "non-automated" requirement in a first time. But I think it is important to have it listed, it is a big requirement to have a cluster with Camel K available.

To automate, several ideas:
- ask , if not exist yet, a `kamel check install` or something like that
- try to use a no-op `kamel` call which returns an error if not available (maybe `kamel get`)
- I often see the error message that "integrations" namespace is missing when Camel k instance is not installed. Maybe we can check for that using kubectl

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,132 @@
...
+      from("timer:java?period=1s")
+        .routeId("java")
+        .setBody()
+          .simple("Hello Camel K from ${routeId}")
+        .to("log:info");
+
+  }
+}
+
+```
```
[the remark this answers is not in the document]
bfitzpat: Whether you see the interest or not, it's staying put. Not everybody is going to use the tools we provide and we need to offer additional methods.

didact/camelk/first-integration.md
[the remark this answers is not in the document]
apupier: ok, was unaware of this convention

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,132 @@
...
+
+While our integration is running in `Dev Mode`, we can modify it and see those changes reflected in the deployed integration. Let's try that now.
+
+Change the message sent to the `.simple()` command of the Camel route in quotes to `We just changed our first Camel K integration while it was running!`. 
+
+Save the file and see what happens in the Output channel.
+
+## Step 5: Managing our Integration
+
+We can see what integrations we currently have running in our Minikube system in the `Apache Camel K Integrations` view in the Explorer activity (Ctrl+Shift+E).
```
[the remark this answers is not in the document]
bfitzpat: Let's go with "the" instead of "what" or "which" and it will be clearer.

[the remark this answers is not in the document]
bfitzpat: Changed to "our environment"

didact/camelk/first-integration.md
```diff
@@ -0,0 +1,132 @@
...
+
+You must have a few things set up prior to walking through the steps in this tutorial. 
+
+<a href='didact://?commandId=vscode.didact.validateAllRequirements' title='Validate all requirements!'><button>Validate all Requirements at Once!</button></a>
+
+| Requirement (Click to Verify)  | Availability | Additional Information/Solution |
+| :--- | :--- | :--- |
+| [At least one folder exists in the workspace](didact://?commandId=vscode.didact.workspaceFolderExistsCheck&text=workspace-folder-status "Ensure that at least one folder exists in the user workspace"){.didact} | *Status: unknown*{#workspace-folder-status} | Create a workspace folder (or [click here to create a temporary folder](didact://?commandId=vscode.didact.createWorkspaceFolder "Create a temporary folder and add it to the workspace."){.didact}), close, and reopen the Didact window
+| [Is Kamel available?](didact://?commandId=vscode.didact.cliCommandSuccessful&text=kamel-status$$kamel "Tests to see if `kamel` returns a result"){.didact} 	| *Status: unknown*{#kamel-status} 	| See [Installing Camel K](https://camel.apache.org/camel-k/latest/installation/installation.html "Documentation on how to Install Apache Camel K")
```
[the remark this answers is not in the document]
bfitzpat: I think it needs to stay put, so I'll leave it.

didact/camelk/first-integration.md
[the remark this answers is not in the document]
bfitzpat: It's been added as a manual step for now

didact/camelk/first-integration.md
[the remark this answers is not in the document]
bfitzpat: No worries. There is no "standard" but I think that Katacoda was what may have started that trend. It uses a "return" image to hint that a link is executable 
image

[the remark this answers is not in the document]
bfitzpat: For now we're using the `Execute^` in a similar fashion until we come up with something better. Nicola started that trend I think

didact/camelk/first-integration.md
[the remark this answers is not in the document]
apupier: why is it needed? The kamel executable on the path of the user is not needed for this tutorial. Or is it checking something else?

src/IntegrationUtils.ts
```diff
@@ -59,13 +60,53 @@ const choiceList = [
...
+
+		if (args && args.length > 0) {
+			// if there are multiple arguments, assume didact input
+			// at this time we only support the URI & basic/dev mode type of integration
+			if (args[0] instanceof vscode.Uri) {
+				context = args[0];
+				inChoice = undefined;
+			} else if (Array.isArray(args[0])) {
+				let innerArgs1 : any[] = args[0];
+				let innerArgs2 : any[] = innerArgs1[0];
```
apupier, line 76: this is considerably complexifying the code. I spent 20 minutes to understand what it is doing.

I don't have an easy solution.
The only little thing that i found that might help is to provide more precise signature, something like: ` export function startIntegration(args: [vscode.Uri | [[vscode.Uri, string?]]]): Promise<boolean> {`

bfitzpat: I cannot change the signature. As I pointed out earlier, regardless of what I put in the function argument declaration, the first argument always comes through as an array of arguments. It appears that execute isn't distinguishing between a context-only argument list and a mix of arguments... it's always passed an array of any[]...

executeCommand<T>(command: string, ...rest: any[]): Thenable<T | undefined>

I have tried multiple ways to make it less complex. It cannot change. The nested series of arrays is crazy, I know.

src/IntegrationUtils.ts
```diff
@@ -509,4 +550,33 @@ export function killChildProcessForIntegration(integration:string) : Promise<boo
...
+		}
+		if (basicIntegration.startsWith(inChoice)) {
+			return basicIntegration;
+		}
+	}
+	// other choices not supported at present because they will require additional inputs
+	return undefined;
+ }
+
+ export async function getCurrentFileSelectionPath(): Promise<vscode.Uri> {
```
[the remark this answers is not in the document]
bfitzpat: Can you clarify what you're saying here?

src/IntegrationUtils.ts
apupier: I tried with `export function startIntegration(args: [vscode.Uri | [[vscode.Uri, string?]]]): Promise<boolean> {` and it seems to work

src/IntegrationUtils.ts
```diff
@@ -509,4 +550,33 @@ export function killChildProcessForIntegration(integration:string) : Promise<boo
...
+	if (inChoice) {
+		if (devModeIntegration.startsWith(inChoice)) {
+			return devModeIntegration;
+		}
+		if (basicIntegration.startsWith(inChoice)) {
+			return basicIntegration;
+		}
+	}
+	// other choices not supported at present because they will require additional inputs
+	return undefined;
```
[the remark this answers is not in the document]
bfitzpat: Because quite honestly that is beyond the scope of this extension's code. There will always be some give and take between Didact and other extensions. Didact can call anything after all and there's no way to protect against that. We may very well add additional support to this code in the future for the other choices, but for now I think this is enough. We can perhaps add some documentation around what is supported for Didact calls, but codifying it is an unneeded step.

didact/camelk/first-integration.md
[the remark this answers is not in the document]
bfitzpat: @apupier , because quite honestly I want it included. End of story.

src/IntegrationUtils.ts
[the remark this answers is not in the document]
apupier: on master:
- have a not-Camel K artefacts (readme.md for instance) opened in editor
- open command palette
- "Start Apache Camel Integration"
- "Start in Dev mode"
- `Starting new integration via Kamel executable. ` is logged in output channel but nothing is deployed
with this PR:
- have a not-Camel K artefacts (readme.md for instance) opened in editor
- open command palette
- "Start Apache Camel Integration"
- "Start in Dev mode"
- the not-Camel K files is tried to be deployed. it is taking times and resources on the cluster to try to start it.

src/IntegrationUtils.ts
bfitzpat: that seems to work for what exactly? that's not any less confusing to me

apupier: it allows to have a signature which is more precise than an array of anything. it is shows that it is supporting an array of vscode.Uri or an array of array of array with first a vscode.Uri and then a potential string. But it is not even true. i don't understand why it was working when i used it.
I'm really confused by this code and the TypeScript notation.

bfitzpat: No, I understand what you were after by changing the signature. I don't disagree that it's confusing. I'm just saying that I tried various combinations and none of them worked consistently. I think that it's a case where you have to delete the out folder and recompile because you're changing the signature -- and that in turn changes the TypeScript to JavaScript conversion...

src/IntegrationUtils.ts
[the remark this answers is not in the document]
bfitzpat: Ok, I think I have this resolved now.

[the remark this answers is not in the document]
bfitzpat: Testing via command palette didn't occur to me. I usually just test via the right-click menu and the didact link, which both worked. Now we do a bit more checking to ensure that things occur correctly. 

Note that you will still see "Starting new integration via Kamel executable" because that message occurs earlier than the startIntegration call itself, but the next line will be `Kamel `execution return code: No valid file specified to start integration function call `

didact/camelk/first-integration.md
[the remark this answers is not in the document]
apupier: advantage to include it:
- <please provide some>
drawbacks:
- longer requirement list
- requires user to download kamel binary
- requires user to add kamel binary on path
- the requirement is not a real one for the tutorial

apart from that. it remains that the comment and pointed link is not about kamel binary installation but global Camel K installation.

src/IntegrationUtils.ts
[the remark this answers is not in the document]
apupier: Given all the complexity added to support the Didact case, it seems important to try to clarify all the things that we can. Fine to handle it in another iteration.

reported https://issues.redhat.com/browse/FUSETOOLS2-471

apupier:
What about adding a test ensuring that the Tutorial remains valid over time?

didact/camelk/first-integration.md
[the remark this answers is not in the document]
bfitzpat: @apupier I'm not playing this game. Thanks. Leave it alone

bfitzpat:
> What about adding a test ensuring that the Tutorial remains valid over time?

Testing is a concern for sure. I need to investigate this from the Didact side of things though, not the Camel K side.

bfitzpat:
@apupier feel free to provide PRs to fix any other issues you want to address

PR #396, opened 2020-04-29
FUSETOOLS2-411 - rework conditional wait and try to have more details in

apupier:
case of failures

it seems to do the trick:
on 6 builds launched on Jenkins all of them are in success for the initially incriminated tests. 1 has a failing test but this would be another flaky tests due to other reasons.

as it is flaky we cannot be sure, so I think that keeping the trace in console log in test execution is better.

see builds 19 to 24 https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/VS%20Code/job/vscode-camelk-release-temporary-testing/

### DIFF
--- a/src/test/suite/Utils.ts
+++ b/src/test/suite/Utils.ts
@@ -22,19 +22,32 @@ import * as vscode from 'vscode';
 const waitUntil = require('async-wait-until');
 
 const extensionId = 'redhat.vscode-camelk';
+export const ACTIVATION_TIMEOUT = 45000;
 
 export async function ensureExtensionActivated() {
 	const extension = vscode.extensions.getExtension(extensionId);
-	if (extension !== null && extension !== undefined) {
-		await waitUntil(() => {
-			return extension.isActive;
-		}, 45000).catch( () => {
-			extension.activate().then( () => {
-				assert.ok("Camel K extension is ready to go");
-			});
-		});
+	if (extension) {
+		await waitInCaseExtensionIsActivating(extension);
+		if(!extension.isActive) {
+			await forceActivation(extension);
+		}
 	} else {
 		assert.fail("Camel K extension is undefined and cannot be activated");
 	}
 	return extension;
+}
+
+async function forceActivation(extension: vscode.Extension<any>) {
+	await extension.activate();
+	await waitUntil(() => {
+		return extension.isActive;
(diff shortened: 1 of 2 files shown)


src/test/suite/Utils.ts
```diff
@@ -28,11 +28,20 @@ export async function ensureExtensionActivated() {
...
-			});
+		}, 45000).catch(() => {
+			console.log('Extension has not started automatically, we will force call to activate it.');
 		});
+		if(!extension.isActive) {
+			await extension.activate();
+			await waitUntil(() => {
+				return extension.isActive;
+			}, 45000).catch(() => {
+				console.log('Extension is not active even after calling activate explicitely.');
```
lhein, line 39: I am wondering if it wouldn't be easier to just fail with that message directly in the catch. It will fail in the next if isActive anyways, so why just do a console.log? What do you think?

bfitzpat: yeah, I kind of agree here. I suspect that we'll end up with a missed failure otherwise

lhein: Only if the extension.isActive out of the sudden will be true. At least that's how I interpret that code passage.

bfitzpat: I read it as...

* explicitly activate the extension
* wait 45000ms to see if it's successful, if not, log a message
* then check it a final time to see if it's successful, if not, fail

I think the timeout should be sufficient. if it's not successful in that timeframe, just fail it

also -- `explicitely` is misspelled -- should be `explicitly`

apupier: > I am wondering if it wouldn't be easier to just fail with that message directly in the catch. It will fail in the next if isActive anyways, so why just do a console.log? What do you think?

I tried and it was failing

apupier: rethinking about it, I think I see why this piece of code is not clear (for you and for the me-one-day-after). I will rework that.

apupier:
I reworked the piece of code. I hope it is clearer now. (at least I find it clearer myself, but maybe I need to wait for the me-one-day-after to check :-) )

I relaunched on Jenkins (with better modified Jenkinsfile to have green status in case of success instead of failing at the publish step). 6 out of 6 job in success. it sounds better.

Ready for a second review

PR #399, opened 2020-05-04
FUSETOOLS2-290 - provide completion for trait properties

apupier:
- range needs to be dummy specified otherwise VS Code framework is
filtering completion out for some reasons (see
https://github.com/microsoft/vscode/issues/95449#issuecomment-621896285
)
- prefix is required to have completions appearing before the long list
of defaults variables in tasks.json context

completiontraitProperty

for next iterations:
- provide default value of property https://issues.redhat.com/browse/FUSETOOLS2-419
- provide snippet with a list of potential value for the property in case of enums https://issues.redhat.com/browse/FUSETOOLS2-420

a non-regression tes for the whole use case, needs ot beprovided at UI level due to mentioned VS Code behavior https://github.com/microsoft/vscode/issues/95449#issuecomment-621896285 , created https://issues.redhat.com/browse/FUSETOOLS2-418

### DIFF
--- a/src/task/CamelKTaskCompletionItemProvider.ts
+++ b/src/task/CamelKTaskCompletionItemProvider.ts
@@ -23,10 +23,10 @@ import { TraitManager } from './TraitManager';
 export class CamelKTaskCompletionItemProvider implements vscode.CompletionItemProvider {
 
 	provideCompletionItems(document: vscode.TextDocument, position: vscode.Position, token: vscode.CancellationToken, context: vscode.CompletionContext): vscode.ProviderResult<vscode.CompletionItem[] | vscode.CompletionList> {
-		return this.provideCompletionItemsForText(document.getText(), document.offsetAt(position));
+		return this.provideCompletionItemsForText(document.getText(), document.offsetAt(position), position);
 	}
 
-	public async provideCompletionItemsForText(text: string, offset: number): Promise<vscode.CompletionItem[]> {
+	public async provideCompletionItemsForText(text: string, offset: number, position: vscode.Position): Promise<vscode.CompletionItem[]> {
 		const globalNode = jsonparser.parseTree(text);
 		const node = jsonparser.findNodeAtOffset(globalNode, offset, false);
 		let completions: vscode.CompletionItem[] = [];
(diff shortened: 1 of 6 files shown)


src/task/CamelKTaskCompletionItemProvider.ts
```diff
@@ -56,6 +60,17 @@ export class CamelKTaskCompletionItemProvider implements vscode.CompletionItemPr
 		return this.isInArray(node)
 			&& this.isSiblingTraitTasks(node);
 	}
+	
+	private isInTraitsArrayMember(node: jsonparser.Node) {
+		let parent = node.parent;
```
lhein, line 65: I would suggest making this a const.

src/task/CamelKTaskCompletionItemProvider.ts
```diff
@@ -56,6 +60,17 @@ export class CamelKTaskCompletionItemProvider implements vscode.CompletionItemPr
...
+		if (parent && this.isInTraitsArray(parent) && node.type === "string") {
+			let value = node.value as string;
```
lhein, line 67: I would suggest making this a const.

src/task/TraitManager.ts
```diff
@@ -35,16 +37,29 @@ export class TraitManager {
 		}
 		return Promise.resolve(completions);
 	}
+	static async provideTraitProperties(traitName: string, position: vscode.Position): Promise<vscode.CompletionItem[]> {
+		const completions: vscode.CompletionItem[] = [];
+		const traitDefs = await TraitManager.retrieveTraitsDefinitions(traitName);
+		for (const property of traitDefs[0].properties) {
```
lhein, line 43: There is also a `foreach` available. It might be easier to use and read when using it to loop the properties. I'd also recommend to specify types if possible.

Something like...
`traitDefs[0].properties.forEach(` (element: <WhatEverTypePropertiesContains>) => {
   // do something with element
});

src/task/TraitManager.ts
```diff
@@ -35,16 +37,29 @@ export class TraitManager {
...
+			const propertyCompletion: vscode.CompletionItem = {
+				label: property.name,
+				sortText: SORT_PREFIX_TO_HAVE_COMPLETIONS_BEFORE_VARIABLES + property.name,
+				range: new vscode.Range(position, position)
+			};
+			completions.push(propertyCompletion);
+		}
+		return Promise.resolve(completions);
```
lhein, line 51: I am confused by this line. An async function should not need to return something via Promise.resolve() but can simply return whatever variable it wants.

apupier: seems both are working. I don't see clear advantages/drawbacks on the 2 possibilities. Changed to return the type directly instead fo the Promise.

src/task/TraitManager.ts
apupier: seems quite equivalent to read in this case.
I think it is useful if treatments are "streamed"

Changed to use forEach, it will be more consistent if we are using "streamed" modifications.

src/task/TraitManager.ts
lhein: I think when in an async function it will do that automatically for you.

src/task/TraitManager.ts
lhein: My goal was just to show you this alternative because I was rather unsure you knew about it.

apupier: ok, thanks

PR #400, opened 2020-05-04
FUSETOOLS2-311 - fix issue when calling "escape" on task list of "Start

apupier:
Apache Camel Integration" list

it is currently logging an error in the VS Code Camel K output. Not sure
that it is needed but it is coherent with the other cases of the "Start
Apache Camel Integration" command when an "escape" is called in one of
the inputbox

### DIFF
--- a/src/IntegrationUtils.ts
+++ b/src/IntegrationUtils.ts
@@ -142,8 +142,13 @@ const choiceList = [
 					// do nothing with config-map or secret
 					break;
 				case vscodeTasksIntegration:
-					await handleDefinedTask(context);
-					resolve();
+					await handleDefinedTask(context).then(() => {
+						resolve();
+					}).catch(onrejected => {
+						reject(onrejected);
+						errorEncountered = true;
+					});
+					return;
 			}
 
 			if (!errorEncountered) {
@@ -176,6 +181,8 @@ async function handleDefinedTask(context: vscode.Uri) {
 			if(camelKTaskToLaunch) {
 				await vscode.tasks.executeTask(camelKTaskToLaunch);
 			}
+		} else {
+			throw new Error('No Camel K Task chosen');
 		}
 	} else {
 		await vscode.window.showInformationMessage('No Camel K Task applicable has been found. You can create one using "Tasks: Open User Tasks" or by creating a tasks.json file in .vscode folder.');
--- a/src/test/suite/IntegrationUtils.test.ts
+++ b/src/test/suite/IntegrationUtils.test.ts
@@ -19,48 +19,70 @@
 import * as sinon from 'sinon';
 import * as vscode from 'vscode';
 import * as IntegrationUtils from './../../IntegrationUtils';
+import { assert } from 'chai';
 import { getDocUri } from './completion.util';
 
-suite("IntegrationUtil tests", function() {
+suite("IntegrationUtil tests", function () {
 
-    let sandbox: sinon.SinonSandbox;
-    let showQuickPickStub: sinon.SinonStub;
-    let executeTaskStub: sinon.SinonStub;
-    
-    setup(() => {
+	let sandbox: sinon.SinonSandbox;
+	let showQuickPickStub: sinon.SinonStub;
+	let executeTaskStub: sinon.SinonStub;
+	let newIntegrationStub: sinon.SinonStub;
+
+	setup(() => {
 		sandbox = sinon.createSandbox();
-        showQuickPickStub = sandbox.stub(vscode.window, 'showQuickPick');
-        executeTaskStub = sandbox.stub(vscode.tasks, 'executeTask');
-	});	
+		showQuickPickStub = sandbox.stub(vscode.window, 'showQuickPick');
+		executeTaskStub = sandbox.stub(vscode.tasks, 'executeTask');
+		newIntegrationStub = sandbox.stub(IntegrationUtils, 'createNewIntegration');
+	});
 
 	teardown(() => {
-        showQuickPickStub.restore();
-        executeTaskStub.restore();
+		showQuickPickStub.restore();
+		executeTaskStub.restore();
+		newIntegrationStub.restore();
 		sandbox.reset();
 	});
 
-    test("ensure listing Camel K task when accessing 'Start Apache Camel Integration'", async function() {
-        showQuickPickStub.onFirstCall().returns(IntegrationUtils.vscodeTasksIntegration);
-        showQuickPickStub.onSecondCall().returns('Test Camel K task');
-        
-        await IntegrationUtils.startIntegration(getDocUri('MyRouteBuilder.java'));
-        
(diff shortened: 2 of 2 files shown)


src/IntegrationUtils.ts
```diff
@@ -142,8 +142,13 @@ const choiceList = [
 					// do nothing with config-map or secret
 					break;
 				case vscodeTasksIntegration:
-					await handleDefinedTask(context);
-					resolve();
+					await handleDefinedTask(context).then(() => {
```
lhein, line 145: That part doesn't look right. You are mixing async/await with promises. Its safer to only use one of it. I'd suggest to switch to something like: 
`
try {
				await handleDefinedTask(context);
				resolve();
			     } catch( (err) => {
				reject(err);
				errorEncountered = true;
			     };
`

apupier: I wanted to stay coherent with the other parts of the function.

What do you think to create a specific refactor task to handle that for the whole file and to handle it in another Pull Request?

src/test/suite/IntegrationUtils.test.ts
```diff
@@ -55,12 +58,24 @@ suite("IntegrationUtil tests", function() {
         showQuickPickStub.onFirstCall().returns(IntegrationUtils.vscodeTasksIntegration);
         showQuickPickStub.onSecondCall().returns(undefined);
         
-        await IntegrationUtils.startIntegration(getDocUri('ADifferentRouteBuilder.java'));
+        await IntegrationUtils.startIntegration(getDocUri('ADifferentRouteBuilder.java')).catch(() => {});
```
lhein, line 61: Better use 
`
try {
    await IntegrationUtils.startIntegration(getDocUri('ADifferentRouteBuilder.java'));
} catch( (err) => {
    // and do something with the catched error.
});

src/test/suite/IntegrationUtils.test.ts
```diff
@@ -55,12 +58,24 @@ suite("IntegrationUtil tests", function() {
...
             ['Start in dev mode Camel K integration opened in active editor'],
             {placeHolder: 'Choose a predefined task'});
         sinon.assert.notCalled(executeTaskStub);
+	});
+	
+	test("ensure no action called on cancel of defined task from 'Start Apache Camel Integration'", async function() {
+        showQuickPickStub.onFirstCall().returns(IntegrationUtils.vscodeTasksIntegration);
+        showQuickPickStub.onSecondCall().returns(undefined);
+        
+        await IntegrationUtils.startIntegration(getDocUri('MyRouteBuilder.java')).then(() => {
```
lhein, line 73: same as the other 2 comments.

src/IntegrationUtils.ts
lhein: You introduced it in this PR, so it would be good to fix it in here.

apupier: I'm not introducing the new notation in this PR.
This way of writing it is used several times in the same method. The idea is to stay coherent with other parts of the code. Even if not optimal it stays more readable than mixing styles inside the same method.

apupier: it is used in the same method at these places:
- https://github.com/camel-tooling/vscode-camelk/blob/master/src/IntegrationUtils.ts#L82
- https://github.com/camel-tooling/vscode-camelk/blob/99fc0f5c0c16c82d17d5b57126dd34092bf6f9dd/src/IntegrationUtils.ts#L94
- https://github.com/camel-tooling/vscode-camelk/blob/99fc0f5c0c16c82d17d5b57126dd34092bf6f9dd/src/IntegrationUtils.ts#L106
- https://github.com/camel-tooling/vscode-camelk/blob/99fc0f5c0c16c82d17d5b57126dd34092bf6f9dd/src/IntegrationUtils.ts#L118
- https://github.com/camel-tooling/vscode-camelk/blob/99fc0f5c0c16c82d17d5b57126dd34092bf6f9dd/src/IntegrationUtils.ts#L130

lhein: But it wasn't in the position you changed. There you introduced it.

image

apupier: I ensured the code to stay coherent

apupier: I created https://issues.redhat.com/browse/FUSETOOLS2-421 and try to get started to refactor to be coherent at the file level, there are some parts I'm really not confident to touch, for instance:

https://github.com/camel-tooling/vscode-camelk/blob/99fc0f5c0c16c82d17d5b57126dd34092bf6f9dd/src/IntegrationUtils.ts#L244-L290

https://github.com/camel-tooling/vscode-camelk/blob/99fc0f5c0c16c82d17d5b57126dd34092bf6f9dd/src/IntegrationUtils.ts#L292-L322

src/test/suite/IntegrationUtils.test.ts
```diff
@@ -19,48 +19,69 @@
...
+		showQuickPickStub.onSecondCall().returns(undefined);
 
-    test("ensure filtering out Camel K task with unrelated target file when accessing 'Start Apache Camel Integration'", async function() {
-        showQuickPickStub.onFirstCall().returns(IntegrationUtils.vscodeTasksIntegration);
-        showQuickPickStub.onSecondCall().returns(undefined);
-        
-        await IntegrationUtils.startIntegration(getDocUri('ADifferentRouteBuilder.java'));
+		try {
+			await IntegrationUtils.startIntegration(getDocUri('ADifferentRouteBuilder.java'));
+		} catch (onrejected) {
```
lhein, line 64: I am confused by the naming of the error variable. It pretends to be some callback function but its just a plain error var. How about renaming it to something more meaningful?

src/IntegrationUtils.ts
lhein: A good rule of thumb is to always try to leave the code cleaner than you found it.

lhein: I am fine if you want to tackle that in a different PR.

src/test/suite/IntegrationUtils.test.ts
apupier: I'm using the naming of the documentation. Seems to be the most coherent. What would be more meaningful?

src/IntegrationUtils.ts
apupier: the effort to do it is important. The risk of regression is high. It will more than double the size of the PR. This is why I really think it should be handled in another PR. That's great that it has been identified.

src/test/suite/IntegrationUtils.test.ts
lhein: What documentation are you referring to?

apupier: argh, it was the then().catch() documentation. Good *catch* :-D

lhein:
approved once we got a green build

PR #401, opened 2020-05-04
Fusetools2 310 fix unable to call kamel when called from right click

apupier:
requires https://github.com/camel-tooling/vscode-camelk/pull/400

### DIFF
--- a/src/test/suite/IntegrationUtils.test.ts
+++ b/src/test/suite/IntegrationUtils.test.ts
@@ -47,7 +47,8 @@ suite("IntegrationUtil tests", function () {
 		showQuickPickStub.onFirstCall().returns(IntegrationUtils.vscodeTasksIntegration);
 		showQuickPickStub.onSecondCall().returns('Test Camel K task');
 
-		await IntegrationUtils.startIntegration(getDocUri('MyRouteBuilder.java'));
+		const status = await IntegrationUtils.startIntegration(getDocUri('MyRouteBuilder.java'));
+		assert.isTrue(status, 'Command should return boolean for success status');
 
 		sinon.assert.calledWith(showQuickPickStub,
 			['Start in dev mode Camel K integration opened in active editor', 'Test Camel K task'],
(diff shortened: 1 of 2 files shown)


src/test/suite/IntegrationUtils.test.ts
```diff
@@ -47,7 +47,8 @@ suite("IntegrationUtil tests", function () {
 		showQuickPickStub.onFirstCall().returns(IntegrationUtils.vscodeTasksIntegration);
 		showQuickPickStub.onSecondCall().returns('Test Camel K task');
 
-		await IntegrationUtils.startIntegration(getDocUri('MyRouteBuilder.java'));
+		const status = await IntegrationUtils.startIntegration(getDocUri('MyRouteBuilder.java'));
+		assert.isOk(status, 'Command should return boolean for success status');
```
lhein, line 51: Wondering if isOk checks for a boolean and why its obfuscated in that way. assert.isTrue or something like that would be easier to understand.

bfitzpat: Chai assert - https://www.chaijs.com/api/assert/#method_isok - seems to look for a 'truthy' condition. 
Truthy - https://developer.mozilla.org/en-US/docs/Glossary/Truthy
Though I think `isTrue` might be easier to grok inline, I wonder if we would have to be more type aware? https://www.chaijs.com/api/assert/#method_istrue

bfitzpat: But I think isOk or isTrue will amount to the same response and this is more of a style question than anything. isTrue may be less vague, but isOk should work.

lhein: As long as status will contain a set boolean value I agree. If it returns undefined or null it will also validate to false which might not be what we want to check for. I am opting for being very strict on the return type by checking for True or False.

apupier: updated to use isTrue.

I was finding isOk easier to read but as it is not a shared point of view and that isTrue is more precise, using isTrue here seems to be better.

thanks!

PR #403, opened 2020-05-05
FUSETOOLS2-303 - open created file in editor

apupier:
note: updated the kamel invoke method to resolve promise on 'close'
event. Otherwise the Promise was never resolved even if the command
ended.

createnewCamelKFileWithEditorOpening

--> not modified changelog as I don't think it is worthy to add a specific line for it given that the new command to create Integration file is already provided

### DIFF
--- a/src/commands/NewIntegrationFileCommand.ts
+++ b/src/commands/NewIntegrationFileCommand.ts
@@ -47,7 +47,9 @@ export async function create() : Promise<void> {
 			if (filename) {
 				const kamelExe = kamel.create();
 				const newFileFullPath: string = computeFullpath(language, workspaceFolder, filename);
-				kamelExe.invoke(`init "${newFileFullPath}"`);
+				await kamelExe.invoke(`init "${newFileFullPath}"`);
+				const textDocument = await vscode.workspace.openTextDocument(newFileFullPath);
+				await vscode.window.showTextDocument(textDocument);
 			}
 		}
 	}
--- a/src/kamel.ts
+++ b/src/kamel.ts
@@ -87,15 +87,21 @@ async function kamelInternal(command: string, devMode: boolean, namespace : stri
 					}
 					resolve(data);
 				});
-				return;
 			}        
 			if (sr.stderr) {
 				sr.stderr.on('data', function (data) {
 					utils.shareMessage(extension.mainOutputChannel, `${data}`);
 					reject(new Error(data));
 				});
-				return;
-			}				
+			}
+			sr.on('close', function (exitCode) {
+				const exitCodeAsString = exitCode.toString();
+				if(exitCode === 0){
+					resolve(exitCodeAsString);
+				} else {
+					reject(new Error(exitCodeAsString));
+				}
+			});
 		}
 	});
 }
--- a/src/test/suite/_completion/completion.java.test.ts
+++ b/src/test/suite/_completion/completion.java.test.ts
@@ -17,8 +17,8 @@
 'use strict';
 
 import * as vscode from 'vscode';
-import { areJavaDependenciesDownloaded } from '../../JavaDependenciesManager';
-import { getDocUri, checkExpectedCompletion } from './completion.util';
+import { areJavaDependenciesDownloaded } from '../../../JavaDependenciesManager';
+import { getDocUri, checkExpectedCompletion } from '../completion.util';
 import { fail } from 'assert';
 
 const os = require('os');
--- a/src/test/suite/_completion/completion.tasks.test.ts
+++ b/src/test/suite/_completion/completion.tasks.test.ts
@@ -17,7 +17,7 @@
 'use strict';
 
 import * as vscode from 'vscode';
-import { getDocUri, checkExpectedCompletion } from './completion.util';
+import { getDocUri, checkExpectedCompletion } from '../completion.util';
 const os = require('os');
 
 suite('Should do completion in tasks.json', () => {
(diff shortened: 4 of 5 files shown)


apupier:
one test is failing on Travis:

```
  1) Should do completion in Camel K standalone files
       Completes from method for Java:
     Error: Timed out after waiting for 10000ms
  	at Timeout.onTimeout (node_modules/async-wait-until/src/waitUntil.js:53:14)
  	at listOnTimeout (internal/timers.js:531:17)
  	at processTimers (internal/timers.js:475:7)
```

I don't get how this test can influence it for now... Will need to investigate. To ease investigation, it works fine locally of course.

apupier:
The failing test for Java completion was occurring on CI only if a Java file is opened before playing the test. I'm unable to reproduce locally. I modified the folder in which the test is part to ensure that it is played at the beginning of the test suite. Unfortunately, i don't have a better solution for now.

src/commands/NewIntegrationFileCommand.ts
```diff
@@ -47,7 +47,9 @@ export async function create() : Promise<void> {
 			if (filename) {
 				const kamelExe = kamel.create();
 				const newFileFullPath: string = computeFullpath(language, workspaceFolder, filename);
-				kamelExe.invoke(`init "${newFileFullPath}"`);
+				await kamelExe.invoke(`init "${newFileFullPath}"`);
```
lhein, line 50: Maybe its worth to try one of the available commands to open a file. 

workbench.action.files.openFile
workbench.action.focusQuickOpen
workbench.action.openEditorAtIndex

bfitzpat, line 50: Would it be better to do `await kamelExe.invoke(...).then() here?

lhein: No. invoke returns a Promise I guess. Await does resolve the result of that promise. I would not use .then afterwards.

apupier: using a command won't break compilation if it is removed/modified/parameters changed.

bfitzpat: vscode.openFolder works great -- just pass it the path, which we already have

bfitzpat: ok. just a thought. :)

apupier: `workbench.action.files.openFile` is opening local file browser to pick a file. (tried with a string and an uri as parameter)
`workbench.action.focusQuickOpen` is opening a local file browser to pick a file when providing an URI and do nothing with a string

also, when using commands, how to know which parameters need sot be provided? Looking into the source code is really nto convenient.

src/test/suite/completion.util.ts
```diff
@@ -27,14 +28,19 @@ export const getDocUri = (p: string) => {
...
-            hasExpectedCompletion = actualCompletionLabelList.includes(expectedCompletion.label);
-        });
-        return hasExpectedCompletion;
-    }, 10000, 500);
+	let hasExpectedCompletion = false;
+	let actualCompletionLabelList;
+	try {
+		await waitUntil(() => {
+			// Executing the command `vscode.executeCompletionItemProvider` to simulate triggering completion
+			(vscode.commands.executeCommand('vscode.executeCompletionItemProvider', docUri, position)).then(value => {
```
lhein, line 36: there shouldn't be a need to put brackets around vscode.commands.xxx. Also I wouldn't mix in Promise.then here but stick with async / await to be more consistent but on first sight everything in the .then() doesn't seem to return a promise.

src/test/suite/completion.util.ts
```diff
@@ -27,14 +28,19 @@ export const getDocUri = (p: string) => {
...
+				let actualCompletionList = value as vscode.CompletionList;
```
lhein, line 37: if value is of type vscode.CompletionList I would define that type directly for value.

src/commands/NewIntegrationFileCommand.ts
bfitzpat: As discussed on our team call, maybe we go with the API approach unless there's good reason to use a command in this case.

src/test/suite/completion.util.ts
bfitzpat: brackets? or parentheses?

src/test/suite/completion.util.ts
```diff
@@ -27,14 +28,19 @@ export const getDocUri = (p: string) => {
...
+				actualCompletionLabelList = actualCompletionList.items.map(c => { return c.label; });
+				hasExpectedCompletion = actualCompletionLabelList.includes(expectedCompletion.label);
+			});
+			return hasExpectedCompletion;
```
lhein, line 41: I am not sure this is correct. It should be chained to the .then (is there a .finally?). Also you forgot a .catch() in case of an exception.  I would really suggest to stick with either async/await or promise chaining.

src/test/suite/completion.util.ts
lhein: damn....parantheses :)

apupier:
remaining comments were on code that was not mandatory for the feature. A part of it was not really modified in this PR and I don't find how to improve it. I removed it to handle it in another PR.

I created another issue to improve this piece of code with details on my investigations on it https://issues.redhat.com/browse/FUSETOOLS2-430

src/test/suite/completion.util.ts
apupier: see https://issues.redhat.com/browse/FUSETOOLS2-430

src/test/suite/completion.util.ts
apupier: see https://issues.redhat.com/browse/FUSETOOLS2-430

src/test/suite/completion.util.ts
apupier: see https://issues.redhat.com/browse/FUSETOOLS2-430

PR #404, opened 2020-05-05
FUSETOOLS2-349 - upgrade kubectl from 1.17.4 to 1.18.2

apupier:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ before_install:
   - 'nvm install --lts'
   - 'export PATH=./node_modules/.bin:$PATH'
   - 'npm install -g typescript'
+  - 'sudo apt-get -qq -y install conntrack'
 install:
   - 'npm install'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -15,6 +15,7 @@ before_install:
   - 'nvm install --lts'
   - 'export PATH=./node_modules/.bin:$PATH'
   - 'npm install -g typescript'
+  - 'sudo apt-get -qq -y install conntrack'
```
lhein, line 18: What is conntrack used for?

bfitzpat: Looks like it's necessary for minikube 1.18 now based on some github issues -- https://github.com/kubernetes/minikube/issues/6619 and https://github.com/kubernetes/minikube/issues/7179 for example

lhein: Ah, thanks for clarification.

PR #407, opened 2020-05-07
FUSETOOLS2-4 - improve message error when Camel K installation not

apupier:
correct

would be better if underlying kamel/kubectl provides more precise
message but I think they are targeting users with different set of
skills and it will require a lot more work to include it upstream. So as
a first iteration,
here is this not very nice but  improving slightly UX for users.

testing it would require quite a lot of refactor and mock. Not sure if
it is worthy.

If the approach seems fine, we can discuss to take time to write the test or not. i wanted to share this kind of modification before investing any time on the testing because even if improving UX slightly, it is really ugly. But I don't have a better idea on short term.

### DIFF
--- a/src/kamel.ts
+++ b/src/kamel.ts
@@ -130,7 +130,19 @@ async function kamelInternalArgs(args: string[], devMode: boolean, namespace: st
 				}        
 				if (sr.stderr) {
 					sr.stderr.on('data', function (data) {
-						utils.shareMessage(extension.mainOutputChannel, `${data}`);
+						const errorMessage = `${data}`;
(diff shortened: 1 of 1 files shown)


bfitzpat:
At first glance, this looks solid. My only concern is how often this could show up. If there's no working container on the system (i.e. no openshift/minikube configured), how many times are those messages going to show up in the output channel? I would like to explore further to make sure we don't just bombard the user with messages unnecessarily (just requires a bit of testing on my part - I will try to do it today).

apupier:
> how many times are those messages going to show up in the output channel?

it should appear every time the user is trying to start an Integration and only in this case.

PR #435, opened 2020-06-08
FUSETOOLS2-472 - remove requirements on kamel CLI

apupier:
kamel CLI on system path is not a requirement to use the tutorial. Also
the provided link and explanation were for general Camel K installation
not for kamel CLI.


lhein:
Can you explain a bit more detailed why kamel CLI is not required? I always thought that `kamel` is used to query available deployed integrations and also to deploy / undeploy and to view the logs.

apupier:
> Can you explain a bit more detailed why kamel CLI is not required? I always thought that `kamel` is used to query available deployed integrations and also to deploy / undeploy and to view the logs.

the kamel version is dynamically downloaded by the VS Code extension when the version set in the settings is not available on the command-line.

lhein:
That would mean it is a requirement?

apupier:
> That would mean it is a requirement?

no. if the re is no kamel binary, it is automatically downloaded by VS Code. The user has nothing to do.

lhein:
It still is a requirement because it wouldn't work if it wouldn't be downloaded for the user automatically. So for me the question would be if its really worth removing it from the requirements list. What happens if the automatic download fails for some reason? The user would not detect that a requirement for the tutorial is not met. It would just fail.

apupier:
> It still is a requirement because it wouldn't work if it wouldn't be downloaded for the user automatically.

why it wouldn't be downloaded? it would mean we have a bug.

> What happens if the automatic download fails for some reason?

there would be an error message

lhein:
> why it wouldn't be downloaded? it would mean we have a bug.

it could be anything. a bug, problem with the internet connection, github being down etc.

> there would be an error message

Yes and that message would be more or less cryptic to the user in comparison to the didact check.

apupier:
> > why it wouldn't be downloaded? it would mean we have a bug.
> 
> it could be anything. a bug, problem with the internet connection, github being down etc.
 
a bug --> we need to fix it
problem with internet connection, github being down --> user will have the same problem to install the kamel CLI when using the tutorial


> > there would be an error message
> 
> Yes and that message would be more or less cryptic to the user in comparison to the didact check.

Why would it be cryptic?
if cryptic, it would be best to improve the error message instead of adding extra work which can be avoided 99% of the time. Even more that currently the pointed description and documentation for that is inaccurate in the tutorial. (created bug report https://issues.redhat.com/browse/FUSETOOLS-3342 to fix it in case we don't remove this invalid requirement)

lhein:
> problem with internet connection, github being down --> user will have the same problem to install the kamel CLI when using the tutorial

Yes, but the main difference is that the tutorial would detect that in the requirements checks and signal that to the user. So it is quite easy to spot.

apupier:
created https://issues.redhat.com/browse/FUSETOOLS2-484 for improvement

PR #436, opened 2020-06-08
FUSETOOLS2-476 - remove non optimal information to copy-paste code

apupier:
- VS Code is providing a command which is more straightforward
- it will avoid to have to maintain the code to copy
- it simplifies the tutorial by providing a single optimal path to work
instead of 2 different paths.

### DIFF
--- a/didact/camelk/first-integration.md
+++ b/didact/camelk/first-integration.md
@@ -49,40 +49,9 @@ Inside VS Code, press `F1` or `Ctrl+Shift+P` to bring up the Command Palette, an
 
 When you complete step 3, you should see the file `Simple.java` appear in your workspace folder. [(Execute^)](didact://?commandId=camelk.integrations.createNewIntegrationFile&text=Simple$$Java)
 
-<details><summary>Advanced Users!</summary>
-
-If you simply want to get started writing some Java, create a file called `Simple.java`, and copy in the following code:
-
-```java
-// camel-k: language=java
-
-import org.apache.camel.builder.RouteBuilder;
-
-public class Simple extends RouteBuilder {
-  @Override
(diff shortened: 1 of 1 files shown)


didact/camelk/first-integration.md
```diff
@@ -49,40 +49,9 @@ Inside VS Code, press `F1` or `Ctrl+Shift+P` to bring up the Command Palette, an
...
-</details>
-
 ## Step 2: Exploring the Java integration file
 
-Now that you have an integration file, let's take a quick look at it. If you created the file yourself, go ahead and open it now. Go to the Explorer activity (Ctrl+Shift+E) and look at the workspace files and folders listed to find `Simple.java`.
-
-- [ ] If you created the file in your workspace earlier by hand, you can open the Simple.java file in the editor.[(Execute^)](didact://?commandId=vscode.openFolder&projectFilePath=Simple.java "Opens the Simple.java file"){.didact}
-
-For this file, we're simply telling Camel to put the message `Hello Camel K from ${routeId}` in the console once a second.
+In the created file in previous step, we're simply telling Camel to put the message `Hello Camel K from ${routeId}` in the console once a second.
```
lhein, line 54: The file we created in the previous step instructs Camel to log the message `Hello Camel K from ${routeId}` in the console once a second.

lhein:
I am a bit undecided if its really that bad to have an alternative way to let users create that file but I can also see the benefits you mentioned.

apupier:
> I am a bit undecided if its really that bad to have an alternative way to let users create that file but I can also see the benefits you mentioned.

can you provide an added-value of the "alternative way" which is ,not even fully described (how to create the empty file?)? For now I see none and found several drawbacks.

PR #437, opened 2020-06-08
FUSETOOLS2-478 - add tutorial information in readme

apupier:


README.md
```diff
@@ -39,6 +39,14 @@ The **Tooling for Apache Camel K** extension is available in the VS Code Extensi
 
     You can install them by following the same steps except search for Kubernetes or Apache Camel in the list of extensions.
 
+## get started using Didact tutorial
```
lhein, line 42: all headlines start with capital letter

README.md
```diff
@@ -39,6 +39,14 @@ The **Tooling for Apache Camel K** extension is available in the VS Code Extensi
...
+
+A tutorial to get started is available when [VS Code Didact](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-didact) is installed. It is accessible through the `Didact Tutorials` Tree view:
```
lhein, line 44: I am not a friend of calling it Tree view. It is on first sight just a view. Tree is an implementation detail users don't really care about.

README.md
```diff
@@ -39,6 +39,14 @@ The **Tooling for Apache Camel K** extension is available in the VS Code Extensi
 
     You can install them by following the same steps except search for Kubernetes or Apache Camel in the list of extensions.
 
+## Get started using Didact tutorial
+
+A tutorial to get started is available when [VS Code Didact](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-didact) is installed. It is accessible through the `Didact Tutorials` view:
+
+![Start tutorial](images/tutorial-access.png)
+
+You can also follow the following steps in this readme. The first steps will be similar. More topics are covered in readme.
```
lhein, line 48: how about 
"You can also follow these steps in this readme. The first steps will be similar but the readme covers more topics."

On a second thought the above statement means that we tell the user to better look at the readme instead of using the didact tutorial because it covers more topics. I would think we should not do that. My suggestions would be to describe only a single "optimal" way here as you suggested in https://github.com/camel-tooling/vscode-camelk/pull/436. So actually I would suggest removing the whole passage "You can also follow the following steps in this readme. The first steps will be similar. More topics are covered in readme."

apupier: I think it is important to mention that the readme contains a broader scope.

To avoid let user thinks that he better have to look at the readme instead of tutorial, what about:
`You can also follow the steps in this readme. The tutorial is providing a finer step-by-step description and is integrated in VS Code. The readme avoids the need to install Didact extension and cover a broader scope.`
or
`The tutorial is providing a detailed step-by-step description and is integrated in VS Code. When finished, you can continue to look to this readme which contains broader scope information.`

lhein: I like the second approach.

PR #438, opened 2020-06-08
FUSETOOLS2-482 - use correct style to emphasize in Didact tutorial

apupier:
it was using style of mentioning UI text or specific naming.

on master:
Screenshot from 2020-06-08 11-51-20
with this PR:
Screenshot from 2020-06-08 11-50-53

### DIFF
--- a/didact/camelk/first-integration.md
+++ b/didact/camelk/first-integration.md
@@ -100,7 +100,7 @@ Since this is likely the first time you've started a new integration in Camel K,
 
 While our integration is running in `Dev Mode`, we can modify it and see those changes reflected in the deployed integration. Let's try that now.
 
-Change the message sent to the `.simple()` command of the Camel route in quotes to `We just changed our first Camel K integration while it was running!`. 
+Change the message sent to the `.simple()` command of the Camel route in quotes to **We just changed our first Camel K integration while it was running!**. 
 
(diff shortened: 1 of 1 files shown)


lhein:
Tbh I don't see the point here. I think that the upper picture makes more sense but I am open to hear you arguments.

apupier:
The text "We just changed our first Camel K integration while it was running!" is not a text that is displayed in UI, neither a specific terminology.
I'm guessing that te intent was to emphasize it as it is revealing one of the biggest feature.
Emphasis is usually done in Markdown using italics or bold https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#emphasis .

The backtick is supposed to be used for "Code and Syntax highlighting", see https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code-and-syntax-highlighting. I don't think there is something specific for UI text or specific terminology, so I often see, as here, to use the backtick for that.

lhein:
If that is the case then I would suggest changing to italic instead and maybe put some enclosing " or ' to make it even more clear. Just bold signals importance to me and not some text to change in the source code. That's why I like the old pic more.

apupier:
ok, I just understood the purpose of this part of text. It is the text to change, not the description of what we are doing.
Current state is the correct one :-)
My bad

PR #440, opened 2020-06-08
FUSETOOLS2-483 - fix kamel binary description of requirement

apupier:
remaining problems:
- this is not a real requirement
- it is checking for kamel available on CLI. it is not looking for
automatically provided one
- cannot point to Camel K documentation as there is no specific section
for kamel binary, it is a subparagraph included in the whole Camel K
installation.


apupier:
no more needed as it was completely removed as it wasn't an accurate requirement

PR #446, opened 2020-06-11
FUSETOOLS2-490 - fix master

apupier:
- use rhel8 insteadof rhel 7 to workaround https://github.com/microsoft/vscode/issues/99492
- use extension API to be able to reach variables used in tests
- deactivate 2 tests in kubectlwatcher that requires too much work of using Object oriented programmation (without insurance that it be enough)

### DIFF
--- a/src/CamelKJSONUtils.ts
+++ b/src/CamelKJSONUtils.ts
@@ -14,53 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import * as fs from 'fs';
 import * as vscode from 'vscode';
-
-const camelAPIVersion = "v1alpha1";
-
-export function stringifyFileContents(absoluteFilePath:string) : Promise<string> {
-	return new Promise( (resolve, reject) => {
-		var text = fs.readFileSync(absoluteFilePath);
-		if (text) {
-			resolve(text.toString());
-		} else {
-			reject(new Error(`Unable to read content of ${absoluteFilePath}.`));
-		}
-	});
-}
-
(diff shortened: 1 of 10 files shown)


src/CamelKJSONUtils.ts
```diff
@@ -14,54 +14,8 @@
...
-			"apiVersion":"camel.apache.org/" + camelAPIVersion,
-			"metadata": {
-				"name" : name.toLowerCase()
-			},
-			"spec" : {
-				"sources" : [
-					{
-						"content" : fileContent,
-						"name" : fileName
-					}
```
apupier, line 46: unused methods which had tests that were failing with VS Code 1.46.0

src/JavaDependenciesManager.ts
```diff
@@ -19,69 +19,66 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 
 const PREFERENCE_KEY_JAVA_REFERENCED_LIBRARIES = "java.project.referencedLibraries";
+export const CAMEL_VERSION = "3.3.0";
 
-export var areJavaDependenciesDownloaded = false;
```
apupier, line 23: this flag is not used anymore. We are using an indirect check, looking at the filesystem. With VS Code 1.46.0, the field was always seen as undefined.

apupier:
the PR job is green on CI https://travis-ci.org/github/camel-tooling/vscode-camelk/builds/698943162

apupier:
> What caused such a huge ripple? New version of TS?

seems new version of VS Code

apupier:
BTW created https://issues.redhat.com/browse/FUSETOOLS2-496 to reactivate the 2 tests that are disabled

PR #455, opened 2020-06-19
FUSETOOLS2-348 - change naming for clarity kamelExe -> kamelExecutor

apupier:


apupier:
> Why not just change it to "kamel" and get rid of the "exe" or "executable" entirely?

the name `kamel` is already used as import https://github.com/camel-tooling/vscode-camelk/pull/455/files#diff-100f87a292f22366ce0098d2e1dd5cd6R22

PR #471, opened 2020-07-09
refactoring versionUtils.ts

lhein:

### DIFF
--- a/src/versionUtils.ts
+++ b/src/versionUtils.ts
@@ -15,15 +15,17 @@
  * limitations under the License.
  */
 'use strict';
+
 import * as vscode from 'vscode';
 import * as extension from './extension';
 import { Errorable, failed } from './errorable';
 import * as config from './config';
 import * as kamelCli from './kamel';
-import { platformString, kamelUnavailableRejection } from './installer';
+import { platformString } from './installer';
 import fetch from 'cross-fetch';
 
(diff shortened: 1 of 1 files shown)


src/versionUtils.ts
```diff
@@ -134,23 +119,18 @@ export async function getLatestCamelKVersion(): Promise<Errorable<string>> {
...
-		await kamelLocal.invoke('version')
-			.then((rtnValue) => {
-				const strArray = rtnValue.split(' ');
-				const version = strArray[strArray.length - 1].trim();
-				console.log(`Apache Camel K CLI (kamel) version returned: ${version}`);
-				resolve(version);
-				return;
-			}).catch(kamelUnavailableRejection(reject));
-	});
+async function checkKamelCLIVersion(): Promise<string> {
```
bfitzpat, line 122: I'm a little concerned about this refactoring and the lack of a catch if the version doesn't exist. I tested it locally and got some funky errors in the console, but nothing conclusive. Can we add a try/catch around it?

lhein: A possible exception is catched in `checkKamelNeedsUpdate` function which should imo make a try catch inside this function obsolete.

lhein: I changed the code a bit to try/catch directly in this function and removed it from the calling one.

PR #475, opened 2020-07-10
refactoring of kubectlutils.ts

lhein:

### DIFF
--- a/src/kubectlutils.ts
+++ b/src/kubectlutils.ts
@@ -35,12 +35,12 @@ export async function getNamedListFromKubernetes(itemType : string, extra? : str
 			if (result && result.stderr) {
 				error = result.stderr;
 			}  
-			return error;
+			return Promise.reject(error);
 		} else if (result) {
 			return result.stdout;
 		}
 	}
-	return 'kubectl not available';
+	return Promise.reject('kubectl not available');
 }
 
 export async function getNamedListFromKubernetesThenParseList(itemType : string, extra? : string): Promise<string[]> {
@@ -114,12 +114,12 @@ export async function getPodsFromKubectlCli() : Promise<string> {
 			if (result && result.stderr) {
 				error = result.stderr;
 			}
-			return error;
+			return Promise.reject(error);
 		} else if (result) {
 			return result.stdout;
 		}
 	}
-	return 'kubectl not available';
+	return Promise.reject('kubectl not available');
 }
 
 export async function getKubernetesVersion(): Promise<string | undefined> {
--- a/src/kubectlutils.ts
+++ b/src/kubectlutils.ts
@@ -17,157 +17,123 @@
 import * as k8s from 'vscode-kubernetes-tools-api';
 import * as config from './config';
 
-export function isKubernetesAvailable(): Promise<boolean> {
-	return new Promise<boolean>( (resolve) => {
-		k8s.extension.kubectl.v1
-			.then( (kubectl) => {
-				resolve(kubectl && kubectl.available);
-			})
-			.catch( err => {
-				resolve(false);
-			});
-	});
+export async function isKubernetesAvailable(): Promise<boolean> {
+	const kubectl: k8s.API<k8s.KubectlV1> = await k8s.extension.kubectl.v1;
+	return kubectl && kubectl.available;
 }
 
-export function getNamedListFromKubernetes(itemType : string, extra? : string): Promise<string> {
-	return new Promise<string>( async (resolve, reject) => {
-		await k8s.extension.kubectl.v1
-			.then( async (kubectl) => {
-				let cmd = `get ${itemType}`;
-				if (extra) {
-					cmd += ` ${extra}`;
-				}
-				if (kubectl && kubectl.available) {
-					return await kubectl.api.invokeCommand(cmd);
-				} else {
-					reject(new Error('Kubernetes not available'));
-				}
-			})
-			.then( (result) => {
-				if (!result || result.code !== 0) {
-					let error = `Unable to invoke kubectl to retrieve ${itemType}`;
-					if (result && result.stderr) {
-						error = result.stderr;
-					}  
-					reject(error);
-				} else if (result) {
-					const splitResults = result.stdout;
-					resolve(splitResults);
-				}
-			})
-			.catch( (err) => reject(err) );
-	});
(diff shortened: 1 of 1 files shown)


src/kubectlutils.ts
```diff
@@ -17,157 +17,126 @@
...
+		const result: k8s.KubectlV1.ShellResult | undefined = await kubectl.api.invokeCommand(cmd);
+		if (!result || result.code !== 0) {
+			let error = `Unable to invoke kubectl to retrieve ${itemType}`;
+			if (result && result.stderr) {
+				error = result.stderr;
+			}  
+			return error;
+		} else if (result) {
+			const splitResults = result.stdout;
+			return splitResults;
```
apupier, line 41: can be returned directly

src/kubectlutils.ts
```diff
@@ -17,157 +17,126 @@
...
+		}
+	}
+	return 'Kubernetes not available';
```
apupier, line 44: Isn't `kubectl` which is not available?

src/kubectlutils.ts
```diff
@@ -17,157 +17,126 @@
...
-			});
-	});
+export async function getNamedListFromKubernetesThenParseList(itemType : string, extra? : string): Promise<string[]> {
+	try {
+		const result = await getNamedListFromKubernetes(itemType, extra);
+		const itemList : string[] = parseShellResult(result);
+		return itemList;
+	} catch (error) {
+		return [error];
+	}
```
apupier, line 54: aren't we losing the information that there is an error? the array of string would seem like there is one elements returned instead of an error?

src/kubectlutils.ts
```diff
@@ -17,157 +17,126 @@
...
+		const result: k8s.KubectlV1.ShellResult | undefined = await kubectl.api.invokeCommand(cmd);
+		if (!result || result.code !== 0) {
+			let error: string = `Unable to invoke kubectl to retrieve pod information`;
+			if (result && result.stderr) {
+				error = result.stderr;
+			}
+			return error;
+		} else if (result) {
+			const splitResults: string = result.stdout;
+			return splitResults;
```
apupier, line 122: can return directly `return result.stdout;``

src/kubectlutils.ts
```diff
@@ -17,157 +17,126 @@
...
+		}
+	}
+	return 'Kubernetes not available';
```
apupier, line 125: isn't `kubectl` which is not available?

src/kubectlutils.ts
```diff
@@ -182,25 +151,17 @@ export async function getKubernetesVersion(): Promise<string | undefined> {
...
 			}
-			return;
-		}).catch ( (error) => {
-			reject(error);
-			return;
-		});
-	});
+		}
+	} catch (error) {
+		console.error(error);
```
apupier, line 164: shouldn't we throw an error?

src/kubectlutils.ts
lhein: I just took what was there before, but you are right.

src/kubectlutils.ts
lhein: yeah, that is probably right. I will change that.

src/kubectlutils.ts
lhein: yep, I just copy/pasted.

src/kubectlutils.ts
lhein: yep

src/kubectlutils.ts
bfitzpat: If kubectl is unavailable, it likely means that the Kubernetes tools extension is unavailable (thus the k8s api). Though it may be more correct to say that 'Kubernetes Tools API is not available' -- to our extension, it amounts to the same thing. We are unable to see into a Kubernetes system because the kubectl API isn't there. Is the user really going to care one way or the other?

lhein: Probably not....it doesn't work in that case...that is all he cares about.

src/kubectlutils.ts
bfitzpat: Yes. I tend to agree. The catch before was...

```
			}).catch( (error) => {
				reject(error);
				return;
			});
```

And the error was at least bubbled up the chain. Maybe if we change the return to Promise<string[]|any> and check the return when we get something back from this method?

src/kubectlutils.ts
bfitzpat: See my earlier comment - but whatever we do we should be consistent. :)

src/kubectlutils.ts
bfitzpat: I agree - similar situation to the one in getNamedListFromKubernetesThenParseList

src/kubectlutils.ts
apupier: it can be OpenShift too.

if what is missing is kubectl, we should tell them that kubectl is missing not that something else is missing.
Even without kubectl, some kamel commands can work and reach the Kubernetes/OpenShift instance.

src/kubectlutils.ts
```diff
@@ -17,157 +17,123 @@
...
+		cmd += ` ${extra}`;
+	}
+	if (kubectl && kubectl.available) {
+		const result: k8s.KubectlV1.ShellResult | undefined = await kubectl.api.invokeCommand(cmd);
+		if (!result || result.code !== 0) {
+			let error = `Unable to invoke kubectl to retrieve ${itemType}`;
+			if (result && result.stderr) {
+				error = result.stderr;
+			}  
+			return error;
```
apupier, line 38: shouldn't we throw an error instead of returning it as a strign like if everything went fine?

src/kubectlutils.ts
```diff
@@ -17,157 +17,123 @@
...
+	const kubectl: k8s.API<k8s.KubectlV1> = await k8s.extension.kubectl.v1;
+	const cmd: string = `get pods`;
+	if (kubectl && kubectl.available) {
+		const result: k8s.KubectlV1.ShellResult | undefined = await kubectl.api.invokeCommand(cmd);
+		if (!result || result.code !== 0) {
+			let error: string = `Unable to invoke kubectl to retrieve pod information`;
+			if (result && result.stderr) {
+				error = result.stderr;
+			}
+			return error;
```
apupier, line 117: shouldn't we throw an error instead of returning it as a string like if everything went fine?

PR #479, opened 2020-07-14
bumping both mocha versions

bfitzpat:


bfitzpat:
One adjustment to the mocha runner got us around the failure in https://github.com/camel-tooling/vscode-camelk/pull/476

src/test/suite/index.ts
```diff
@@ -36,15 +36,14 @@ function loadCoverageRunner(testsRoot: string): CoverageRunner | undefined {
 	}
 }
 
-let mocha = new Mocha({
+// Create the mocha test
+const mocha = new Mocha({
 	ui: 'tdd',
 	timeout: 20000,
 	reporter: 'mocha-jenkins-reporter',
-	useColors: true
```
lhein, line 43: regarding https://github.com/mochajs/mocha/pull/4178 we can still use `colors` instead of `useColors`.

bfitzpat: good to know. I wasn't sure what to do with it, but useColors definitely threw an error

bfitzpat: looks like it's color instead of colors, but I'm trying it here

bfitzpat: yup. that worked.

PR #482, opened 2020-07-16
Refactor kubectl

lhein:


lhein:
based on #475

src/kubectl.ts
```diff
@@ -147,43 +153,35 @@ async function kubectllInternalArgs(args: string[], namespace: string | undefine
...
 		env: {
 			HOME: process.env.HOME,
 			PATH: process.env.PATH
 		}
 	};
 
-	const execResult = await shell.execCore(cmd, opts);
-	if (execResult.code) {
-		return { err: execResult.code, output: execResult.stderr };
+	const execResult: ShellResult = await shell.execCore(cmd, opts);
```
lhein, line 182: tbh I am still not 100% sure I understand the original if clause. Imo a result code of `0` is signaling a normal execution of the process while other codes indicate issues. So the question for me would be why any code is handled as an "error" case instead of just the ones that return a code !== 0.

lhein: Ok. I learned now that 0 will result in a falsy statement, so the original code might be fine. I still would keep the change for a better readability.

src/kubectl.ts
```diff
@@ -147,16 +153,35 @@ async function kubectllInternalArgs(args: string[], namespace: string | undefine
...
+			PATH: process.env.PATH
+		}
+	};
+
+	const execResult: ShellResult = await shell.execCore(cmd, opts);
+	if (execResult.code === 0) {
+		return { err: null, output: execResult.stdout };
+	}
+	return { err: execResult.code, output: execResult.stderr };
+}
```
apupier, line 187: I think you had an issue with conflict resolution during rebase. This part has been moved to the shell class

src/kubectl.ts
```diff
@@ -147,16 +153,35 @@ async function kubectllInternalArgs(args: string[], namespace: string | undefine
 }
 
 export async function baseKubectlPath(): Promise<string> {
-	let result : shell.FindBinaryResult = await shell.findBinary('kubectl');
+	const result : FindBinaryResult = await findBinary('kubectl');
```
apupier, line 156: I think you had an issue with conflict resolution during rebase. This part is available in the shell class

src/kubectl.ts
lhein: crap. good catch.

PR #485, opened 2020-07-20
refactored kamel.ts and fixed some errors caused by incorrect catches

lhein:

### DIFF
--- a/src/kamel.ts
+++ b/src/kamel.ts
@@ -21,7 +21,7 @@ import * as utils from './CamelKJSONUtils';
 import * as extension from './extension';
 import * as path from 'path';
 import * as fs from 'fs';
-import { findBinary } from './kubectl';
+import { FindBinaryResult, findBinary } from "./shell";
 
 export interface Kamel {
 	devMode : boolean;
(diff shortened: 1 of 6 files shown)


src/kamel.ts
```diff
@@ -14,15 +14,14 @@
...
-import { exec, spawn} from "child_process";
+import { exec, spawn, ChildProcess} from "child_process";
 import * as child_process from "child_process";
 import * as config from './config';
 import * as utils from './CamelKJSONUtils';
 import * as extension from './extension';
 import * as path from 'path';
 import * as fs from 'fs';
-import * as shell from './shell';
+import { findBinary } from './kubectl';
```
apupier, line 24: That's good to remove duplicated code.
I think the `findBinary` should be extracted to its own class. There is no reason that kamel has a dependency on kubectl.

lhein:
> can you precise which "errors are fixed" that were caused by "incorrect catches"?

see https://travis-ci.org/github/camel-tooling/vscode-camelk/builds/709054714 
lines 645-718

src/kamel.ts
lhein: agreed. moved it to the shell.ts

lhein:
waiting for @bfitzpat  's review before merging

Issue #500, opened 2020-07-30
camelk is downloaded on every start


lhein:

lhein:
the problem is that `kamel version` returns 1.1.0 but the versionToUse is `v1.1.0`.

PR #501, opened 2020-07-30
fixed broken camelK version check due to new tag naming logic #500

lhein:

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
 					},
 					"camelk.integrations.runtimeVersion": {
 						"type": "string",
-						"description": "Default Apache Camel K Runtime Version (equivalent to GitHub tag)",
+						"description": "Default Apache Camel K Runtime Version (for example 1.1.0)",
 						"scope": "window"
 					},
 					"camelk.integrations.autoUpgrade": {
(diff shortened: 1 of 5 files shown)


lhein:
I tried to find the least intrusive solution to this issue. We distinguish tag names and versions now and if the version does not start with 0.x or 1.0 then we assume the tag name prepends a `v` like for `v1.1.0` which introduced this new naming logic. Please test carefully @bfitzpat

bfitzpat:
I will take a look at this in a moment, but I'm curious if it's necessary. When asked, Claus had me create https://github.com/apache/camel-k/issues/1645 so they could avoid the situation moving forward?

bfitzpat:
Ahhh. I was not aware of https://github.com/apache/camel-k/pull/1603

bfitzpat:
A few comments... since the new versions will ALWAYS now start with a v, I like the new approach. 

When I turn off auto upgrade and change the version to v1.1.0, I see `Setting for Apache Camel K runtime version changed to invalid version v1.1.0. Please set the version to a valid runtime version. ` So my minimal validation is working a little at least.

Wondering if we should check for the 'v' in the `applyDefaultVersionSettings` function in extension.ts?

And we should now change the runtime version setting label to remove the tag comment

image

PR #506, opened 2020-08-11
attempting to do more optional chaining

bfitzpat:

### DIFF
--- a/src/test/suite/_completion/completion.java.test.ts
+++ b/src/test/suite/_completion/completion.java.test.ts
@@ -65,11 +65,11 @@ async function testCompletion(
 	let javaExtension: vscode.Extension<any> | undefined;
 	await waitUntil(() => {
 		javaExtension = vscode.extensions.getExtension('redhat.java');
-		return javaExtension?.isActive && javaExtension?.exports.status === "Started";
+		return javaExtension?.isActive && javaExtension?.exports?.status === "Started";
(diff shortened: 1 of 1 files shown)


bfitzpat:
Merging to try and fix the issue that only appears on Jenkins

apupier:
when issue appears only on Jenkins, it is faster and avoids to push changes in master changes by modifying https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/VS%20Code/job/vscode-temporary-testing/ and the jenkins file

bfitzpat:
> when issue appears only on Jenkins, it is faster and avoids to push changes in master changes by modifying https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/VS%20Code/job/vscode-temporary-testing/ and the jenkins file

We should add such details somewhere so we don't lose this information. I will never remember it otherwise

apupier:
> > when issue appears only on Jenkins, it is faster and avoids to push changes in master changes by modifying https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/VS%20Code/job/vscode-temporary-testing/ and the jenkins file
> 
> We should add such details somewhere so we don't lose this information @apupier . I will never remember it otherwise

created https://github.com/jboss-fuse/tooling_material/pull/58

PR #507, opened 2020-08-11
increasing timeout to see if that helps

bfitzpat:


bfitzpat:
Again, since we have a green build I am hoping this helps get us past the Jenkins issue upstream

PR #521, opened 2020-08-20
FUSETOOLS2-607 - add information to update Che registry after VS Code

apupier:
release

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -8,7 +8,7 @@
     * Wait for PR to be merged
 * Check that someone listed as _submitter_ in Jenkinsfile is available
 * Create a tag
-* Push the tag to vscode-vscode repository, it will trigger a build after few minutes
+* Push the tag to vscode-camelk repository, it will trigger a build after few minutes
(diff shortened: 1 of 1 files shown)


lhein (approved):

Contributing.md
```diff
@@ -21,6 +21,10 @@
   * Run 'npm install' so that the package-lock.json is updated
   * Push changes in a PR
   * Follow PR until it is approved/merged
+  
+# Next steps after release
+
+It is highly recommended to integrate the newly released version of VS Code Camel K in Eclipse Che registry. See [Contributing.md in Che registry](https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/vscode-camelk/Contributing.md).
```
bfitzpat, line 27: Language can be improved a little here. I would get rid of "it is highly recommended", since all of these are just steps in a process and this should definitely be done if the rest of the steps are done too. 

"Create a PR to integrate the newly released version of VS Code Camel K in the Eclipse Che Registry."

apupier: it can be several PRs if it requires an update of the sidecar or not.

PR #522, opened 2020-08-21
refactored NewIntegrationFileCommand.ts

lhein:

### DIFF
--- a/src/commands/NewIntegrationFileCommand.ts
+++ b/src/commands/NewIntegrationFileCommand.ts
@@ -33,7 +33,6 @@ const LANGUAGES_WITH_FILENAME_EXTENSIONS = new Map([
 const LANGUAGES = Array.from(LANGUAGES_WITH_FILENAME_EXTENSIONS.keys());
 
 export async function create(...args: any[]) : Promise<void> {
-
 	let language : string | undefined;
 	let workspaceFolder : vscode.WorkspaceFolder | undefined;
 	if (vscode.workspace.workspaceFolders) {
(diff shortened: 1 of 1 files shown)


lhein:
that happened automatically when doing `npm install`. I think in an earlier commit this hasn't been done after updating the package.json and that's why it now has a change.

apupier:
> that happened automatically when doing `npm install`. I think in an earlier commit this hasn't been done after updating the package.json and that's why it now has a change.

ok so unrelated to this PR. better to do in a separate PR

lhein:
This is such a minor thing that its not really worth the hassle at all.

apupier:
> This is such a minor thing that its not really worth the hassle at all.

8000+ modifications of lines doesn't sound minor to me

lhein:
And you would review those 8000+ lines if they were in a separate PR, would you? :)

apupier:
> And you would review those 8000+ modifications if they were in a separate PR, would you? :)

I would be able to start discussion about this specific change. Why there are so many changes? Are there ways to avoid big bang like that? more easily be able to compare the result with a local call myself of the "npm install".
These are questions unrelated to the "refactored newIntegrationFileCommand" that I consequently not asked

it will also allow easier browsing of the history/dichotomized potential bug research in case one or the other introduced a problem.

lhein:
depending on #526

lhein:
once #526 has been merged I will rebase this PR

apupier:
I recommend to modify the branch to not be based on the modification of the other PR as they are unrelated.

lhein:
Once the rebase is done the changed package-lock will automatically disappear from this PR as far as I understand.

apupier:
> Once the rebase is done the changed package-lock will automatically disappear from this PR as far as I understand.

you will have to potentially fix conflicts.
Until the other PR is merged, it is then blocking the review and merge of this PR. Having a branch with separated concerns when they don't really depend on each other allows to merge one or the other in whatever order.

PR #526, opened 2020-08-24
updated package lock triggered by npm install on clean master

lhein:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,11 +13,172 @@
 				"@babel/highlight": "^7.10.4"
 			}
 		},
+		"@babel/core": {
+			"version": "7.11.4",
+			"resolved": "https://registry.npmjs.org/@babel/core/-/core-7.11.4.tgz",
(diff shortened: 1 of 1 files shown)


apupier:
Why are there so many changes? isn't dependabot supposed to upgrade dependencies? Are there ways to avoid big bang like that?

lhein:
Ask npm :)

apupier:
if I run locally, I end up with the same files than on current master. See https://github.com/lhein/vscode-camelk/pull/1

lhein:
Closing this PR as we get no agreement for a merge.

apupier:
reasons provided for not merging:
- Avoid the file to have thousands of modified lines on each PR. For instance, it is easier to avoid having a package-lock.json that might point to an alternative npm registry manager that can contain malicious code.
- No need to commit of package-lock.json until we understand what is going on sounds like a good thing as no Pull request or planned work (as far as I know) is blocked by that.
It will be a problem only when you add a dependency or update one manually without relying on dependabot.

As nothing is blocked, there is time to investigate on this various track if needed:
- google search leads to find old bugs leading to differences when stuff is called with [empty node_modules ](https://npm.community/t/npm-install-changes-package-lock-extensively/7653/3)for instance
- What are the changes more precisely?
- raise a question on npm support
- can it be reproduced on other OS than ArchLinux?
- Are there global npm configurations that can have an impact on that?

PR #530, opened 2020-08-26
FUSETOOLS2-571 - provide setting to change the default Camel K schema association

apupier:
settingToChangeCamelKAssociatedSchema


apupier:
pushed on the wrong repository

PR #531, opened 2020-08-26
FUSETOOLS2-571 - provide setting to change the default Camel K schema

apupier:
association

settingToChangeCamelKAssociatedSchema

### DIFF
--- a/src/CamelKSchemaManager.ts
+++ b/src/CamelKSchemaManager.ts
@@ -21,9 +21,10 @@ import fetch from 'cross-fetch';
 
 const CAMELK_MODELINE_PREFIX = '# camel-k:';
 const CAMELK_SCHEMA_ID = 'camelk';
-const CAMELK_SCHEMA_URI_PREFIX = CAMELK_SCHEMA_ID + '://schema/';
+export const CAMELK_SCHEMA_URI_PREFIX = CAMELK_SCHEMA_ID + '://schema/';
 
-let camelkSchemaCache: string;
+let camelkSchemaCache: string | undefined;
+let configUpdaterRegistered: boolean = false;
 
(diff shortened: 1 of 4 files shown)


src/CamelKSchemaManager.ts
```diff
@@ -34,20 +35,40 @@ export async function registerCamelKSchemaProvider(mainOutputChannel: vscode.Out
...
+		vscode.workspace.onDidChangeConfiguration(event => {
+			if (event.affectsConfiguration('camelk.yaml.schema')) {
+				retrieveSchemaAsCache(mainOutputChannel);
+			}
+		});
+	}
 }
 
 async function retrieveSchemaAsCache(mainOutputChannel: vscode.OutputChannel): Promise<string | undefined> {
+	camelkSchemaCache = undefined;
```
lhein, line 54: Just for clarification: If we currently have some schema set in the schemaCache and the user sets it to an invalid URI, then the old cached schema is gone. I am wondering if we should be more graceful and only change the schemaCache if the download / retrieval succeeds and if not display some error notification. What do you think?

apupier: I think we should not keep the old one as it means that there will be the wrong schema provided compared to the value that the user is providing in the setting.

Keeping the old schema in cache would avoid another external call. To benefit from that and not hitting the problem of incoherence between the setting and the schema used, it can be possible to keep a map of already retrieved schema. I felt it was overkill for the functionality. This can be improved later on anyway.

If the download/retrieval failed, there is currently a message in the output log. Having a validation in VS Code settings would be nice. Also possible to provide a notification but it might be a bit too intrusive for the feature. Both can be improved later on.

lhein (approved):

PR #540, opened 2020-09-09
FUSETOOLS2-613 - Provide test starting an integration

apupier:
- start with one DSL kind (Java) and basic deployment to simplify
- timeout is quite long, it is related to the time required by first
Camel K Integration to be started as it is downloading a bunch of
dependencies
- it is checking the deployment through the Camel K integration view
explorer

~~one important point missing might be to remove the deployed integration at the end of the test. Not a problem for Travis Ci as it is the only one used currently but will be necessary when increasing coverage~~


src/test/suite/integrationExplorer.test.ts
```diff
@@ -41,6 +41,7 @@ suite('Camel-k Integrations View', () => {
 
 	teardown(() => {
 		sandbox.restore();
+		integrationExplorer.setRetrieveIntegrations(true);
```
apupier, line 44: important to avoid side-effect of having Integration explorer effectively showing integrations in the explorer

apupier:
nota: created https://issues.redhat.com/browse/FUSETOOLS2-671 to handle side effect caused by extension.test.ts in another PR. This is not related to this PR, it has just revealed it.

bfitzpat (approved):
Interesting. Will we eventually add other types of tests (not just Java, but maybe XML and YAML perhaps)?

src/test/suite/integrationExplorer.test.ts
bfitzpat: Good catch!

bfitzpat:
> nota: created https://issues.redhat.com/browse/FUSETOOLS2-671 to handle side effect caused by extension.test.ts in another PR. This is not related to this PR, it has just revealed it.

This is an interesting twist. And it's caused by the order of tests?

apupier:
> > nota: created https://issues.redhat.com/browse/FUSETOOLS2-671 to handle side effect caused by extension.test.ts in another PR. This is not related to this PR, it has just revealed it.
> 
> This is an interesting twist. And it's caused by the order of tests?

changing the order of test avoid to hit the side effect. The side effect still exists, just hidden.

apupier:
> Will we eventually add other types of tests (not just Java, but maybe XML and YAML perhaps)?

the goal is to cover all functionalities. So yes that and also test the various ways to start an integration

PR #547, opened 2020-09-17
Fusetools2 572 fixname space

apupier:


apupier:
- adding a test starting an integration in a specific namespace
- fix refresh of Integration explorer view when updating namespace (detected while writing the test and required to write the test)

src/kamel.ts
```diff
@@ -41,7 +41,7 @@ class KamelImpl implements Kamel {
 	}
 
 	async getPath(): Promise<string> {
-		return await baseKamelPath();
+		return (await baseKamelPath()).replace('\n', '');
```
bfitzpat, line 44: Do we know where the \n was slipping into the path?

bfitzpat (approved):
Looks good - just wondering about the `\n`

src/kamel.ts
apupier: if I understood well, it is the "which" command that is used. the output is adding a \n to have the prompt going to a new line.

apupier: wonderign if not better to replace in the baseKamelPath method then

bfitzpat: Yeah, that's kind of what I was thinking too -- might be better to move up stream a bit further?

bfitzpat (approved):

src/kamel.ts
apupier: agree. done

PR #558, opened 2020-09-23
FUSETOOLS2-712 - upgrade minikube to 1.13.1 on Travis infra

apupier:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_script:
   # Download and provide in path kubectl , which is a requirement for using minikube.
   - curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.18.2/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
   # Download and provide in path minikube.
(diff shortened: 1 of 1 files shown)


bfitzpat (changes requested):
Why are we upgrading? Is it merely a version bump or is there a more specific reason? There is no justification for the bump in the JIRA or the PR, so it's without context

apupier:
> Why are we upgrading? Is it merely a version bump or is there a more specific reason? There is no justification for the bump in the JIRA or the PR, so it's without context

it is to fill technical debt.
it also ensures to be working with latest available version.
Dependabot is not able to update automatically as this is not a regular dependency.

bfitzpat:
> it is to fill technical debt.
> it also ensures to be working with latest available version.
> Dependabot is not able to update automatically as this is not a regular dependency.

This would be good to include in the JIRA so when our QE folks look at it they have some context to base it on. :)

apupier:
> > it is to fill technical debt.
> > it also ensures to be working with latest available version.
> > Dependabot is not able to update automatically as this is not a regular dependency.
> 
> This would be good to include in the JIRA so when our QE folks look at it they have some context to base it on. :)

done

bfitzpat (approved):

lhein (approved):

PR #564, opened 2020-09-29
FUSETOOLS2-674 - add automated test of basic deployment for all DSLs

apupier:
tests results can be seen here: https://travis-ci.org/github/camel-tooling/vscode-camelk/builds/731349784#L662-L700

### DIFF
--- a/src/test/suite/StartIntegration.test.ts
+++ b/src/test/suite/StartIntegration.test.ts
@@ -28,6 +28,7 @@ import { getNamedListFromKubernetesThenParseList } from '../../kubectlutils';
 import * as shelljs from 'shelljs';
 import * as kamel from './../../kamel';
 import * as kubectl from './../../kubectl';
+import { LANGUAGES, LANGUAGES_WITH_FILENAME_EXTENSIONS } from '../../commands/NewIntegrationFileCommand';
 
 const RUNNING_TIMEOUT: number = 360000;
 const DEPLOYED_TIMEOUT: number = 5000;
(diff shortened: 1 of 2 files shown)


bfitzpat:
What does this line "Removed the child process running in the background for test-basic-java-script-deploy: false" mean? The output in the test console is a little confusing for me.

apupier:
> What does this line "Removed the child process running in the background for test-basic-java-script-deploy: false" mean? The output in the test console is a little confusing for me.

this has been introduced a while back https://github.com/camel-tooling/vscode-camelk/blame/43f8806b88107a44dd08da2d824359c88063d3d3/src/extension.ts#L115

it is logged when an integration is removed. If I understand well, it return "false" when no "watch" process which i used only in dev mode.

apupier:
> > What does this line "Removed the child process running in the background for test-basic-java-script-deploy: false" mean? The output in the test console is a little confusing for me.
> 
> this has been introduced a while back https://github.com/camel-tooling/vscode-camelk/blame/43f8806b88107a44dd08da2d824359c88063d3d3/src/extension.ts#L115
> 
> it is logged when an integration is removed. If I understand well, it return "false" when no "watch" process which i used only in dev mode.

well, no tin this case, there is one that is "true" but no integration is launched in dev mode https://travis-ci.org/github/camel-tooling/vscode-camelk/builds/731349784#L673

apupier:
> > > What does this line "Removed the child process running in the background for test-basic-java-script-deploy: false" mean? The output in the test console is a little confusing for me.
> > 
> > 
> > this has been introduced a while back https://github.com/camel-tooling/vscode-camelk/blame/43f8806b88107a44dd08da2d824359c88063d3d3/src/extension.ts#L115
> > it is logged when an integration is removed. If I understand well, it return "false" when no "watch" process which i used only in dev mode.
> 
> well, no tin this case, there is one that is "true" but no integration is launched in dev mode https://travis-ci.org/github/camel-tooling/vscode-camelk/builds/731349784#L673

created https://issues.redhat.com/browse/FUSETOOLS2-727 to improve this log message which was not touched in this PR.

src/test/suite/StartIntegration.test.ts
```diff
@@ -66,18 +67,22 @@ suite('Check can deploy default examples', () => {
...
-
-	const testJava = test('Check can deploy Java example', async () => {
-		skipOnJenkins(testJava);
-		createdFile = await createFile(showQuickpickStub, showWorkspaceFolderPickStub, showInputBoxStub, 'TestBasicDeploy');
-
-		await startIntegrationWithBasicCheck(showQuickpickStub);
-	}).timeout(TOTAL_TIMEOUT);
+	
+	suite('Check basic deployments for each languages', function() {
+		LANGUAGES.forEach(function(language) {
```
bfitzpat, line 72: @lhein wasn't there an issue with using foreach and async operations? or am I imagining that?

apupier: this is what official Mocha documentation is using https://mochajs.org/#dynamically-generating-tests

apupier: I remember Lars mentioning that:
```
- `return` in a `forEach` loop doesn't return 😄  
   --> see https://stackoverflow.com/questions/2641347/short-circuit-array-foreach-like-calling-break
```
isn't this one? or it is something else that you have in mind?

lhein: https://codeburst.io/javascript-async-await-with-foreach-b6ba62bbf404

apupier: ok so we are not in this case. The direct function inside the foreach is not async

bfitzpat (approved):

PR #570, opened 2020-10-06
FUSETOOLS2-222 - support task also when kamel is not on user path

apupier:
test covering this use case provided in this PR https://github.com/camel-tooling/vscode-camelk/pull/571

### DIFF
--- a/src/task/CamelKTaskDefinition.ts
+++ b/src/task/CamelKTaskDefinition.ts
@@ -16,6 +16,7 @@
  */
 'use strict';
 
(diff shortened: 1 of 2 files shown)


apupier:
current failure is due to the rate limit reached to access github. Github tokens are not used in Travis on PR branches that are outside of the main repository...

bfitzpat (approved):
Makes sense you'd have to wait for tasks to complete

lhein (approved):

PR #608, opened 2020-11-17
FUSETOOLS2-878 - update link to Camel K schema

apupier:
we were pointing to master as it was implemented when there was no tag
available yet. it was refactored on master. So pointing to latest tag.
It will avoid breakage from external part but requires more maintenance.

including more traces in log too for next time, it will be easier to
debug (see https://travis-ci.com/github/camel-tooling/vscode-camelk/builds/202096761#L391 for instance)

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
 					},
 					"camelk.yaml.schema": {
 						"type": "string",
-						"default": "https://raw.githubusercontent.com/apache/camel-k-runtime/master/camel-k-loader-yaml/camel-k-loader-yaml/src/generated/resources/camel-yaml-dsl.json",
+						"default": "https://raw.githubusercontent.com/apache/camel-k-runtime/camel-k-runtime-parent-1.5.0/camel-k-loader-yaml/camel-k-loader-yaml/src/generated/resources/camel-yaml-dsl.json",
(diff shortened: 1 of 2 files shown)


bfitzpat (approved):

package.json
```diff
@@ -96,7 +96,7 @@
 					},
 					"camelk.yaml.schema": {
 						"type": "string",
-						"default": "https://raw.githubusercontent.com/apache/camel-k-runtime/master/camel-k-loader-yaml/camel-k-loader-yaml/src/generated/resources/camel-yaml-dsl.json",
+						"default": "https://raw.githubusercontent.com/apache/camel-k-runtime/camel-k-runtime-parent-1.5.0/camel-k-loader-yaml/camel-k-loader-yaml/src/generated/resources/camel-yaml-dsl.json",
```
lhein, line 99: Now we won't detect problems in newer versions right away. Is that wanted behavior?

apupier: it will avoid to break on already released version.

There won't be undetected breakage as it will be detected when the url is updated.

We are not testing the content of the json schema. This extension is not responsible for it. We are testing that we can retrieve the json schema.

lhein (approved):

PR #652, opened 2021-01-25
FUSETOOLS2-954 - due to Minikube wrong latest .deb distributed, specify

apupier:
a specific version

see https://github.com/kubernetes/minikube/issues/10224


apupier:
merged in combined PR https://github.com/camel-tooling/vscode-camelk/pull/653

PR #661, opened 2021-02-09
FUSETOOLS2-985 - fix master

apupier:
- improve message on error
- increase tiemout, since VS Code 1.53.0, it seems that VS Code Java is takign more time to activate


src/test/suite/_completion/completion.java.test.ts
```diff
@@ -65,11 +65,22 @@ async function testCompletion(
...
-		return javaExtension?.isActive && javaExtension?.exports?.status === "Started";
+		const isJavaExtensionActive = javaExtension?.isActive;
+		if(isJavaExtensionActive) {
+			try {
+				return javaExtension?.exports?.status === "Started";
+			} catch (error) {
+				console.log(error);
+				return false;
+			}
+		} else {
```
lhein, line 76: the else case is not really required. a simple `return false;` would be sufficient. wdyt?

apupier: I think it is more readable to have clearly the if/else

lhein (approved):

bfitzpat (approved):

PR #698, opened 2021-03-19
FUSETOOLS2-1033 telemetry integration creation

apupier:
https://snapshot.woopra.com/5GeHozdVXv4ujHQiSigFTx4oQrhfF6NsCMMikBkSmgQ3xNLcyuouWzs2BScnU6EGi8

### DIFF
--- a/src/test/runTests.ts
+++ b/src/test/runTests.ts
@@ -16,6 +16,7 @@ async function main() : Promise<void> {
 		const cliPath: string = resolveCliPathFromVSCodeExecutablePath(vscodeExecutablePath);
 		installExtraExtension(cliPath, 'ms-kubernetes-tools.vscode-kubernetes-tools');
 		installExtraExtension(cliPath, 'redhat.java');
+		installExtraExtension(cliPath, 'redhat.vscode-commons');
 
 		await runTests({ vscodeExecutablePath, extensionDevelopmentPath, extensionTestsPath, launchArgs: [testWorkspace] });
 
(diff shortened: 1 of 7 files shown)


apupier:
removed the test part.
it was passing locally because I activated the opt-in feature.
In theory, I tried to spy the TelemetryService.send method which is called by our codebase but it seems it was spying the one from vscode-redhat-telemetry. And unfortunately not found a reliable way to spy ours. Created this branch for my attempt https://github.com/apupier/vscode-camelk/tree/FUSETOOLS-1033-telemetryIntegrationCreation-testAttempt
I think we can go without for now. It won't affect end users even if the events are no more sent, we will "just" lose the metrics for us.

src/test/runTests.ts
```diff
@@ -16,6 +16,7 @@ async function main() : Promise<void> {
 		const cliPath: string = resolveCliPathFromVSCodeExecutablePath(vscodeExecutablePath);
 		installExtraExtension(cliPath, 'ms-kubernetes-tools.vscode-kubernetes-tools');
 		installExtraExtension(cliPath, 'redhat.java');
+		installExtraExtension(cliPath, 'redhat.vscode-commons');
```
bfitzpat, line 19: Do we still need this if there's no tests?

apupier: yes, the redhat.vscode-commons is declared as an `extensionDependencies`. Conssequently, to have the VS Code Camel k starting during the tests, this extension needs to be installed.

bfitzpat (approved):

lhein (approved):

PR #703, opened 2021-03-22
FUSETOOLS2-1033 - Provide test for telemetry of Integration file

apupier:
creation

extracting the code in a specific file allows to spy the correct instance/method.


lhein (approved):

bfitzpat (approved):
Very cool! Glad you found a way to test in a bit more granular fashion.

PR #704, opened 2021-03-22
FUSETOOLS-1035 - provide basic telemetry for most commands

apupier:
- there is no existing test at top level of commands call. I have not
found how
to add an assertion without high complexity
- it is covering only "happy" case. In case of failure, no telemetry
reported in most case. In case of abortion of the command, a telemetry
event is reported without mentioning that user didn't go to the end of
the "wizard command"

https://snapshot.woopra.com/5GeHozdVXv4ujHQiSgQLcHBerREYPXSpGkCqLoJkd5rBWPKq3qjREgUCrjqDCdDqR3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the "vscode-camelk" extension will be documented in this 
 
 ## 0.0.22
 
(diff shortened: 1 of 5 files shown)


bfitzpat (approved):
Definitely!

lhein (approved):

PR #718, opened 2021-04-01
FUSETOOLS2-1043 - update compilation classpath when ready

apupier:
https://drive.google.com/file/d/1ukRoit7eG1DqxE4NiJO2DC4js8XTRBRa/view?usp=sharing

it allows to have compilation classpath well detected when isntalling VS
Code Camel K extension and that the Java Camel K file is already open.

as it is a complicated use case to test (need to uninstall extension
under test, clean settings folder or restart vs code and then reinstall
it), I do not know how it can be easily automated at the level we have
setup the test infrastructure.

for the future, might be nice to have a feedback to the user on the
progress of the downloads so that he/she knows that the "current
compilation" error should vanish soon. But my last attempt to have
feedback progress from Maven launched through node was a failure.


lhein (approved):

bfitzpat (approved):
Nice!

PR #735, opened 2021-04-26
FUSETOOLS2-1066 - switching to latest Camel K version 1.4.0

lhein:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ jobs:
           name: Configure Kamel
           command: |               
             # Download and provide in path kamel.
-            curl -Lo kamel.tar.gz https://github.com/apache/camel-k/releases/download/v1.3.2/camel-k-client-1.3.2-linux-64bit.tar.gz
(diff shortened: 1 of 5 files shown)


apupier (changes requested):
- the version of Camel dependencies needs to be updated from 3.7.0 to 3.9.0 
https://github.com/camel-tooling/vscode-camelk/blob/9f8ad157c9c76de099dd345b0265f389a5bad043/src/JavaDependenciesManager.ts#L22
https://github.com/apache/camel-k-runtime/blob/7a0629cfcad4a295196e9392e478d3153290258a/pom.xml#L41
- better to update minikube in another commit as it is unrealted to Camel K 1.4.0 (unless it requires this updated new version?)

lhein:
Thanks Aurelien. I missed that Camel version bump.

apupier (changes requested):
- please add an entry in teh changelog
- I think we should wait for camel language server to support the new way of wrting mvn dependency in modeline to merge this one

lhein:
yes. good points.

apupier (approved):

PR #742, opened 2021-04-28
FUSETOOLS2-1036 - provide more precise telemetry for Start integration

apupier:
command

- language of file by providing the extension; prefer to provide the extension to avoid to handle all potential cases of extension yaml/yml/YAML and surely other ways to write that I miss)
- main option of the command

it will give this kind of reports:

Screenshot from 2021-04-28 20-43-39
Screenshot from 2021-04-28 20-43-18

### DIFF
--- a/src/test/suite/StartIntegration.test.ts
+++ b/src/test/suite/StartIntegration.test.ts
@@ -26,10 +26,13 @@ import { assert, expect } from 'chai';
 import { waitUntil } from 'async-wait-until';
 import { getNamedListFromKubernetesThenParseList } from '../../kubectlutils';
 import * as shelljs from 'shelljs';
+import * as extension from '../../extension';
 import * as kamel from './../../kamel';
 import * as kubectl from './../../kubectl';
 import { LANGUAGES, LANGUAGES_WITH_FILENAME_EXTENSIONS } from '../../commands/NewIntegrationFileCommand';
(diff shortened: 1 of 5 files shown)


src/test/suite/StartIntegration.test.ts
```diff
@@ -44,16 +46,18 @@ suite('Check can deploy default examples', () => {
...
 		showQuickpickStub = sinon.stub(vscode.window, 'showQuickPick');
 		showInputBoxStub = sinon.stub(vscode.window, 'showInputBox');
 		showWorkspaceFolderPickStub = sinon.stub(vscode.window, 'showWorkspaceFolderPick');
 		// Workaround due to bug in shelljs: https://github.com/shelljs/shelljs/issues/704
 		shelljs.config.execPath = shelljs.which('node').toString();
+		telemetrySpy = sinon.spy(await getTelemetryServiceInstance(), 'send');
 	});
 
-	teardown(() => {
+	teardown(async() => {
```
lhein, line 60: why is it required to make it async?

apupier: I thougth it was needed as I switched the setup parameter method to async. And it is nto the case. I switch back to sync as it was before for the teardown which doesn't require it.

lhein (approved):

bfitzpat (approved):
Nice!

PR #745, opened 2021-04-29
FUSETOOLS2-1094 - provide more precise error message in test failure for

apupier:
expected completion not found

the error will be something like:

```
  7) Should do completion in yaml Camel K files without file name pattern
       Completes for first level:
     Error: Error: Timed out after waiting for 10000 ms
Cannot found expected completion "test" in the list of completions:
error-handler
from
my-added-property-for-the-demo
on-exception
rest
route
```

instead of somethign like:
```
  1) Should do completion in Camel K standalone files
       Completes additional dependencies:
     Error: Timed out after waiting for 10000 ms
  	at /home/circleci/project/node_modules/async-wait-until/dist/index.js:1:1034
 ```

### DIFF
--- a/src/test/suite/completion.util.ts
+++ b/src/test/suite/completion.util.ts
@@ -26,18 +26,30 @@ export const getDocUri = (p: string) => {
 };
 
 export async function checkExpectedCompletion(docUri: vscode.Uri, position: vscode.Position, expectedCompletion: vscode.CompletionItem) {
-    let hasExpectedCompletion = false;
-    await waitUntil(() => {
-        // Executing the command `vscode.executeCompletionItemProvider` to simulate triggering completion
-        (vscode.commands.executeCommand('vscode.executeCompletionItemProvider', docUri, position)).then(value => {
-			let actualCompletionList = value as vscode.CompletionList;
-			const completionItemFound = actualCompletionList.items.find(completion => {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by apupier** [comment]
might help to understand why it is failing mostly all the time on this other PR https://github.com/camel-tooling/vscode-camelk/pull/743 (and often on others but not reproduced on this one)

**Review by lhein — APPROVED** [review]

File: src/test/suite/completion.util.ts [hunk 1]
```diff
@@ -26,18 +26,30 @@ export const getDocUri = (p: string) => {
...
+			return hasExpectedCompletion;
+		}, 10000, 500);
+	} catch (err) {
+		let errorMessage = '';
+		if(lastCompletionList) {
+			lastCompletionList.items.forEach(completion => {
+				errorMessage += completion.label + '\n';
+			});
+		}
+		throw new Error(`${err}\nCannot found expected completion "${expectedCompletion.label}" in the list of completions:\n${errorMessage}`);
```
bfitzpat, line 53: Can we change `Cannot found` to `Cannot find`?

bfitzpat (changes requested):
Language should change slightly. "found" vs "find"

bfitzpat (approved):

PR #752, opened 2021-05-06
FUSETOOLS2-944 - provide specific task to activate Camel K debug

apupier:
https://user-images.githubusercontent.com/1105127/117299250-9b50bb80-ae78-11eb-8c65-beecac4c0cb6.mp4

this is a first iteration which avoids to go back to command line for
most cases. Even if it requires several config to be setup first. It
also has the benefit to allow to investigate possibilities more easily
for better support.

- it might allow to combine with a Java Debug launch config
- the parameter for integrationName might need to be different and take
a file name and convert? or both?
- might be better to have a specific launch config instead of task (but
sounds far more complicated)
- maybe can create the camel k debug task and the java debug launch config programmatically to bind altogether on a right-click for instance. Very close to have it done for the test. To be investigated in next iteration


src/task/CamelKDebugTaskDefinition.ts
```diff
@@ -0,0 +1,88 @@
...
+		return new vscode.Task(definition,
+			vscode.TaskScope.Workspace,
+			displayName,
+			CamelKDebugTaskProvider.DEBUG_CAMELK_TYPE,
+			processExecution);
+	}
+}
+
+async function computeKamelDebugCommandLine(definition: CamelKDebugTaskDefinition) {
+	//TODO: put some logic to determine the integration name based on a provided file? it will allow something more dynamic.
```
apupier, line 76: yes or no, i'm still not sure what will the best. it will depend a bit on the path for next iteration. But dicovering the VS Code API, iI already changed my mind several times on the subject. SO Iguess it it fine to do it this way for now.

src/test/suite/Utils/DeployTestUtil.ts
```diff
@@ -0,0 +1,142 @@
...
+import { assert, expect } from 'chai';
+import { waitUntil } from 'async-wait-until';
+import * as extension from '../../../extension';
+import * as kamel from '../../../kamel';
+import { LANGUAGES_WITH_FILENAME_EXTENSIONS } from '../../../commands/NewIntegrationFileCommand';
+import { TelemetryEvent } from '@redhat-developer/vscode-redhat-telemetry/lib';
+import { TreeNode } from '../../../CamelKNodeProvider';
+import { UNDEPLOY_TIMEOUT, PROVIDER_POPULATED_TIMEOUT, RUNNING_TIMEOUT, DEPLOYED_TIMEOUT, EDITOR_OPENED_TIMEOUT } from '../StartIntegration.test';
+
+export async function cleanDeployedIntegration() {
```
apupier, line 31: this clas is an extract of previously existing functions

src/test/suite/DebugIntegration.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+async function getActiveTerminalOutput() : Promise<string> {
+	const term = vscode.window.activeTerminal;
+	console.log(`-current terminal = ${term?.name}`);
+	await vscode.commands.executeCommand('workbench.action.terminal.selectAll');
+	await vscode.commands.executeCommand('workbench.action.terminal.copySelection');
+	await vscode.commands.executeCommand('workbench.action.terminal.clearSelection');	
+	const clipboard_content = await vscode.env.clipboard.readText();
+	return clipboard_content.trim();
+}
```
apupier, line 137: using similar than in VS Code didact

lhein:
That are a lot of manual steps to have the debug enabled. I am wondering if we can somehow further automate this to improve the usability.

bfitzpat:
So... end result = cool. Definitely concerned about the number of manual steps. Even if there was a way to template some of it with comments or feedback to the user indicating what they had to find/copy in, it would be a bit better. Great first steps though!

bfitzpat (approved):

PR #759, opened 2021-05-10
Fusetools2 950 contexttual menu start java debugger

apupier:
~requires https://github.com/camel-tooling/vscode-camelk/pull/752/~

video for simple case: https://youtu.be/TQasRr1PiYw

scope and direct limitations:
- basic use case handled fo rlocalhost only (for remote https://issues.redhat.com/browse/FUSETOOLS2-1122 )
- multiple debug session launched in parallel handled
- contextual menu is always available and enable on right-click on Integrations (https://issues.redhat.com/browse/FUSETOOLS2-1126 )
- command is visible in command palette https://issues.redhat.com/browse/FUSETOOLS2-1127

### DIFF
--- a/src/commands/StartJavaDebuggerCommand.ts
+++ b/src/commands/StartJavaDebuggerCommand.ts
@@ -34,16 +34,19 @@ export async function start(integrationItem: TreeNode): Promise<void> {
 	kamelArgs.push(`${port}`);
 	const childProcess = await kamelExecutor.invokeArgs(kamelArgs);
 	let debuggerLaunched = false;
+	let isListeningFromTransportMessageSent = false;
 	childProcess.stdout?.on('data', function (data) {
 		const messageData: string = `${data}`;
-		if (!debuggerLaunched && messageData.includes('Listening for transport dt_socket at address:')) {
+		console.log(messageData);
+		isListeningFromTransportMessageSent ||= messageData.includes('Listening for transport dt_socket at address:');
+		if (!debuggerLaunched && isListeningFromTransportMessageSent && messageData.includes('Forwarding from')) {
 			const workspaceFolderList = vscode.workspace.workspaceFolders;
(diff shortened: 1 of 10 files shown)


src/commands/StartJavaDebuggerCommand.ts
```diff
@@ -0,0 +1,81 @@
...
+	const port = await retrieveFreeLocalPort();
+	kamelArgs.push(`--port`);
+	kamelArgs.push(`${port}`);
+	const childProcess = await kamelExecutor.invokeArgs(kamelArgs);
+	let debuggerLaunched = false;
+	let isListeningFromTransportMessageSent = false;
+	let isForwardingFrommessageSent = false; 
+	childProcess.stdout?.on('data', function (data) {
+		const messageData: string = `${data}`;
+		isListeningFromTransportMessageSent ||= messageData.includes('Listening for transport dt_socket at address:');
```
lhein, line 41: I think this is making it more complicated and harder to read than it could.

lhein (approved):

src/commands/StartJavaDebuggerCommand.ts
apupier: I find the use of a condition more complicated. It contains a negated value and a branch.

Using the binary operator also requires to read 2 different arguments which requires to check that the second one is the same to which it is reassigned or not.

The logical assignment operator is part of all languages that I know ([Java](https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.22.2), [javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR_assignment), TypeScript, [Rust](https://doc.rust-lang.org/reference/expressions/operator-expr.html#compound-assignment-expressions). It allows to describe the purpose with a direct flow.

PR #761, opened 2021-05-12
FUSETOOLS2-736 - fixed possible exception case when tagName parameter…

lhein:
… is undefined. Also removed doubled call to stripLeadingV

### DIFF
--- a/src/versionUtils.ts
+++ b/src/versionUtils.ts
@@ -96,11 +96,11 @@ export async function pingGithubUrl(urlStr: string): Promise<boolean> {
 }
 
 export function isOldTagNaming(tagName: string): boolean {
-	return tagName.startsWith('0.') || tagName.startsWith('1.0.');
+	return tagName?.startsWith('0.') || tagName?.startsWith('1.0.');
 }
 
 export function stripLeadingV(tagName: string): string {
-	return tagName.startsWith('v') ? tagName.substring(1) : tagName;
+	return tagName?.toLowerCase().startsWith('v') ? tagName.substring(1) : tagName;
 }
 
 export async function getLatestCamelKVersion(): Promise<Errorable<string>> {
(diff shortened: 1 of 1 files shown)


src/versionUtils.ts
```diff
@@ -96,11 +96,11 @@ export async function pingGithubUrl(urlStr: string): Promise<boolean> {
 }
 
 export function isOldTagNaming(tagName: string): boolean {
-	return tagName.startsWith('0.') || tagName.startsWith('1.0.');
+	return tagName?.startsWith('0.') || tagName?.startsWith('1.0.');
```
apupier, line 99: if tagName can be undefined, shouldn't be better to declare it in the parameter as such?

src/versionUtils.ts
```diff
@@ -147,7 +147,7 @@ async function checkKamelCLIVersion(): Promise<string | void> {
 }
 
 export async function handleChangeRuntimeConfiguration() {
-	const runtimeSetting: string = stripLeadingV(config.getKamelRuntimeVersionConfig() as string);
+	const runtimeSetting: string = config.getKamelRuntimeVersionConfig() as string;
```
apupier, line 150: isn"t the root issue caused by this cast to tring although it can be undefined?

lhein: Actually it is. The call to getKamelRuntimeVersionConfig can return undefined so the cast to string is a bad thing.

src/versionUtils.ts
lhein: I just double checked the calling functions and it seems we will not have that case. Anyways having this additional check will not hurt.

bfitzpat: There's a difference between saying a parameter is optional (i.e. `(tagName? : string)`) and checking the value of the parameter passed in. But maybe it could be a bit more precise with `(tagName : string | undefined)` and a check below for a more easily spotted result if the incoming value is undefined? 

That said, I think the way Lars has it now gets the result we're looking for (i.e. false) for undefined values without all the extra work.

src/versionUtils.ts
```diff
@@ -96,11 +96,11 @@ export async function pingGithubUrl(urlStr: string): Promise<boolean> {
 }
 
 export function isOldTagNaming(tagName: string): boolean {
```
bfitzpat, line 98: What does this do if tagName is undefined coming in?

lhein: I check it and it will not happen for now. All the calling functions have a check in place before calling this function OR hand over a fixed string value. This is of course just the current state which can change in future. If the tagName is undefined we will return false here which is not really nice but will fail in the calling method because we would search for a tag named 'v' :)

bfitzpat (approved):

PR #767, opened 2021-05-17
Remove redundant use of `await` on a return value

apupier:
as reported as code smell by Sonar
https://sonarcloud.io/project/issues?id=vscode-camelk&resolved=false&rules=typescript%3AS4326&types=CODE_SMELL

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -160,7 +160,7 @@ export function getKamelRuntimeVersionConfig(): string | undefined {
 
 export async function setKamelRuntimeVersionConfig(value : string) : Promise<void> {
 	const configuration = vscode.workspace.getConfiguration();
-	return await configuration.update(RUNTIME_VERSION_KEY, stripLeadingV(value), vscode.ConfigurationTarget.Global);
+	return configuration.update(RUNTIME_VERSION_KEY, stripLeadingV(value), vscode.ConfigurationTarget.Global);
 }
 
 export function getKamelAutoupgradeConfig() : boolean {
@@ -171,5 +171,5 @@ export function getKamelAutoupgradeConfig() : boolean {
 // for testing purposes 
 export async function setKamelAutoupgradeConfig(value : boolean) : Promise<void> {
 	const configuration = vscode.workspace.getConfiguration();
-	return await configuration.update(AUTOUPGRADE_KEY, value, vscode.ConfigurationTarget.Global);
+	return configuration.update(AUTOUPGRADE_KEY, value, vscode.ConfigurationTarget.Global);
 }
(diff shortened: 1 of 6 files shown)


src/config.ts
```diff
@@ -160,7 +160,7 @@ export function getKamelRuntimeVersionConfig(): string | undefined {
 
 export async function setKamelRuntimeVersionConfig(value : string) : Promise<void> {
 	const configuration = vscode.workspace.getConfiguration();
-	return await configuration.update(RUNTIME_VERSION_KEY, stripLeadingV(value), vscode.ConfigurationTarget.Global);
+	return configuration.update(RUNTIME_VERSION_KEY, stripLeadingV(value), vscode.ConfigurationTarget.Global);
```
bfitzpat, line 163: If await is no longer needed, do we need to return a Promise?

src/config.ts
```diff
@@ -171,5 +171,5 @@ export function getKamelAutoupgradeConfig() : boolean {
 // for testing purposes 
 export async function setKamelAutoupgradeConfig(value : boolean) : Promise<void> {
 	const configuration = vscode.workspace.getConfiguration();
-	return await configuration.update(AUTOUPGRADE_KEY, value, vscode.ConfigurationTarget.Global);
+	return configuration.update(AUTOUPGRADE_KEY, value, vscode.ConfigurationTarget.Global);
```
bfitzpat, line 174: If await is no longer needed, do we need to return a Promise?

src/kamel.ts
```diff
@@ -41,7 +41,7 @@ class KamelImpl implements Kamel {
 	}
 
 	async getPath(): Promise<string> {
-		return await baseKamelPath();
+		return baseKamelPath();
```
bfitzpat, line 44: If await is no longer needed, do we need to return a Promise?

src/kubectl.ts
```diff
@@ -39,7 +39,7 @@ class KubectlImpl implements Kubectl {
 	}
 
 	async getPath(): Promise<string> {
-		return await baseKubectlPath();
+		return baseKubectlPath();
```
bfitzpat, line 42: If await is no longer needed, do we need to return a Promise?

bfitzpat (commented):
So for all of these methods where we're getting rid of await, do we need to return a Promise at all?

src/config.ts
apupier: the `await` is not needed for returned values inside the `async` method which is returning a `Promise`.

more information in the description of the issue on Sonar: https://sonarcloud.io/organizations/camel-tooling/rules?open=typescript%3AS4326&rule_key=typescript%3AS4326

bfitzpat (approved):

PR #768, opened 2021-05-17
Use `for-of` loop instead of a `for` loop

apupier:
as recommended by Sonar
https://sonarcloud.io/project/issues?id=vscode-camelk&open=AXe5MjsIomOWeNW5TYmX&resolved=false&rules=typescript%3AS4138&types=CODE_SMELL

### DIFF
--- a/src/CamelKNodeProvider.ts
+++ b/src/CamelKNodeProvider.ts
@@ -150,13 +150,13 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 		if (output) {
 			const lines: Array<string> = output.split('\n');
 			for (let entry of lines) {
-				const line: Array<string> = entry.split('  ');
+				const spaceSplittedLine: Array<string> = entry.split('  ');
 				const cleanLine: Array<string> = new Array<string>();
-				for (var i=0; i < line.length; i++) {
-					if (line[i].trim().length === 0) {
+				for (let integration of spaceSplittedLine) {
+					if (integration.trim().length === 0) {
 						continue;
 					}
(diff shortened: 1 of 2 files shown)


src/CamelKNodeProvider.ts
```diff
@@ -150,13 +150,13 @@ export class CamelKNodeProvider implements vscode.TreeDataProvider<TreeNode> {
 		if (output) {
 			const lines: Array<string> = output.split('\n');
 			for (let entry of lines) {
-				const line: Array<string> = entry.split('  ');
+				const spaceSplittedLine: Array<string> = entry.split('  ');
 				const cleanLine: Array<string> = new Array<string>();
-				for (var i=0; i < line.length; i++) {
-					if (line[i].trim().length === 0) {
+				for (let integration of spaceSplittedLine) {
```
bfitzpat, line 155: Does entry.split always return a value? Or do we need to watch for undefined/empty array here?

apupier: it always returns an array:

```
    /**
     * Split a string into substrings using the specified separator and return them as an array.
     * @param splitter An object that can split a string.
     * @param limit A value used to limit the number of elements returned in the array.
     */
    split(splitter: { [Symbol.split](string: string, limit?: number): string[]; }, limit?: number): string[];
```

bfitzpat: so the `for-of` construct doesn't execute if the array is empty?

apupier: yes, if the array is empty, it is not entering the statement block.

bfitzpat (approved):

PR #148, opened 2019-09-25
issue #50 - working on downloading the camel k cli if not available

bfitzpat:
* used download-tarball package to retrieve gzipped cli from url, ungzip, and put file in user file system
* check for existence of kamel cli and compare versions to see if we need to install the new one
* check for versions of kubectl and minikube
* created wrapper for kamel cli to handle using kamel on the path or the one we download
* use wrapper where possible
* changed variable name from runKamel to runKubectl in getIntegrationsFromCamelK method to avoid confusion
* ensure that kamel cli is on system path or downloaded to user directory before downloading again
* added a similar wrapper for the kubectl cli that uses the downloaded version in the Kubernetes extension if it's not already on the system path
* removed the Rest Proxy options (will be replaced with Namespace support soon)
* Cleaned up all shell usage so we now only use the kamel, kubectl, and kubernetes API wrappers
* added a test to ensure that we have access to the kamel & kubectl CLIs after installation and that they are executable

Signed-off-by: Brian Fitzpatrick <bfitzpat@redhat.com>

PR #802, opened 2021-06-22
FUSETOOLS2-1168 - Avoid override on startup of the kamel binary when

apupier:
auto-upgrade is off and provided kamel version is not a community
released version

### DIFF
--- a/src/test/suite/versionUtils.test.ts
+++ b/src/test/suite/versionUtils.test.ts
@@ -18,69 +18,108 @@
 
 import * as chai from 'chai';
 import * as sinonChai from 'sinon-chai';
+import * as config from '../../config';
 import { Platform } from '../../shell';
 import * as versionUtils from '../../versionUtils';
+import path = require('path');
 
 chai.use(sinonChai);
 const should = chai.should();
 
-suite("ensure version url methods are functioning as expected", () => {
+suite("VersionUtils check", () => {
 
-	test("validate url for existing 1.0.0 version", async () => {
-		await validateVersion('1.0.0', Platform.LINUX, 'https://github.com/apache/camel-k/releases/download/1.0.0/camel-k-client-1.0.0-linux-64bit.tar.gz');
-	});
+	suite("ensure version url methods are functioning as expected", () => {
 
-	test("validate url for existing 1.0.1 version", async () => {
-		await validateVersion('1.0.1', Platform.LINUX, 'https://github.com/apache/camel-k/releases/download/1.0.1/camel-k-client-1.0.1-linux-64bit.tar.gz');
-	});
+		test("validate url for existing 1.0.0 version", async () => {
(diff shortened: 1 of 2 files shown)


src/test/suite/versionUtils.test.ts
```diff
@@ -18,69 +18,101 @@
...
+		async function invalidateVersion(tagName: string, platform: Platform): Promise<void> {
+			try {
+				await versionUtils.getDownloadURLForCamelKTag(tagName, platform);
+				should.fail(`Downloading invalid Camel-K version (${tagName}) did not fail!`);
+			} catch (error) {
+				should.exist(error);
+			}
 		}
-	}
+	});
```
apupier, line 91: all this part is just that it has been grouped inside a "sub" test suite

apupier:
the problem with current state is the case: auto-upgrade off, no valid tools path provided : it doesn't trigger a download

src/versionUtils.ts
```diff
@@ -47,39 +47,44 @@ export async function testVersionAvailable(versionToUse: string): Promise<boolea
...
-			return false;
-		}
-	} else if (!versionToUse) {
-		const latestversion = await getLatestCamelKVersion();
-		if (failed(latestversion)) {
-			console.error(latestversion.error);
-			return true;
+	const runtimeVersionSetting = config.getKamelRuntimeVersionConfig();
+	const autoUpgradeSetting: boolean | undefined = config.getKamelAutoupgradeConfig();
+	const kamelCliToolsPath: string | undefined = config.getActiveKamelconfig();
```
lhein, line 52: Just a minor thing but in the 2 lines above you are using the term `Setting` in the variable name while in this line you don't. Why not using the same terminology all over?

src/versionUtils.ts
```diff
@@ -47,39 +47,44 @@ export async function testVersionAvailable(versionToUse: string): Promise<boolea
...
+		if (versionToUse) {
+			const passedVersionAvailable = await testVersionAvailable(versionToUse);
+			if (!passedVersionAvailable) {
+				const msg: string = `Camel K CLI Version ${versionToUse} unavailable.`;
+				extension.shareMessageInMainOutputChannel(msg);
+				return false;
+			}
+		} else if (!versionToUse) {
+			const latestversion = await getLatestCamelKVersion();
+			if (failed(latestversion)) {
```
lhein, line 64: I am not a friend of naming a function `failed`. It does say nothing about what the function does.

apupier: This was [introduced a while ago](https://github.com/camel-tooling/vscode-camelk/pull/148) and not modified in this PR. i reported https://issues.redhat.com/browse/FUSETOOLS2-1175

src/test/suite/versionUtils.test.ts
apupier: I recommend to use "hide whitespace change" to hide this kind of change when reviewing this PR

lhein (approved):

bfitzpat (approved):

PR #811, opened 2021-07-02
FUSETOOLS2-676 - provide test for deploy with ConfigMap

apupier:
very basic test, only checking that the integration is running, not
checking that the configmap can be effectively used. Created https://issues.redhat.com/browse/FUSETOOLS2-1196 to test further


src/IntegrationUtils.ts
```diff
@@ -265,17 +265,14 @@ async function handleDefinedTask(context: vscode.Uri) {
...
 					placeHolder: 'Select an available Kubernetes ConfigMap or ESC to cancel'
-				}).then ( (selectedConfigMap) => {
-					resolve(selectedConfigMap);
-				});
-			}).catch ( (error) => {
-				reject(new Error(`No ConfigMaps available: ${error}`));
-			});
-		});
+				}));
+	});
```
apupier, line 275: this modification simplifies a bit the code and seems to be required to be able to mock showQuickPick

lhein (approved):

PR #821, opened 2021-07-05
Fusetools2 1196 provide assertion on what is available

apupier:
based on https://github.com/camel-tooling/vscode-camelk/pull/811

PR #822, opened 2021-07-06
Fusetools2 679 and FUSETOOLS2-1191 check start integration with property and remove the wrong double-quotes

apupier:
based on https://github.com/camel-tooling/vscode-camelk/pull/821

done the FUSETOOLS2-679 and FUSETOOLS2-1191 in the mean time, as writing a test which is checking a wrong behavior doesn't make a lot of sense

startIntegrationWithProperty

### DIFF
--- a/src/test/suite/StartIntegration.test.ts
+++ b/src/test/suite/StartIntegration.test.ts
@@ -40,6 +40,8 @@ export const PROVIDER_POPULATED_TIMEOUT: number = 20000;
 export const EDITOR_OPENED_TIMEOUT: number = 5000;
 const TOTAL_TIMEOUT: number = RUNNING_TIMEOUT + DEPLOYED_TIMEOUT + EDITOR_OPENED_TIMEOUT + UNDEPLOY_TIMEOUT + PROVIDER_POPULATED_TIMEOUT;
 
+const lineReturnAndSpaces: RegExp = /\r?\n|\r|\s/g;
+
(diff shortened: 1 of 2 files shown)


src/test/suite/StartIntegration.test.ts
```diff
@@ -112,6 +156,29 @@ suite('Check can deploy default examples', () => {
...
+	const lineReturnAndSpaces: RegExp = /\r?\n|\r|\s/g;
+	expect(description.replace(lineReturnAndSpaces, '')).includes('Configuration:Type:configmapValue:my-configmap');
+}
+
+async function checkPropertyAvailableAvailableForDeployedIntegration() {
+	const describeShell = shelljs.exec(`${await kamel.create().getPath()} describe integration test-java-deploy-with-property`);
+	const description: string = describeShell.stdout;
+	const lineReturnAndSpaces: RegExp = /\r?\n|\r|\s/g;
+	expect(description).includes('propertyKey=my Value');
+	expect(description.replace(lineReturnAndSpaces, '')).includes('Configuration:Type:propertyValue:propertyKey=myValue');
```
apupier, line 171: an even better check would be to evaluate properties in pod context or to create a route using the property and logging, then to check the log of the pod. But it seems to complicate a lot the test, not sure that it is worthy. it can be done anyway in another iteration.

bfitzpat (approved):
Definitely cleans up that method a bit more. Good work!

PR #823, opened 2021-07-06
FUSETOOLS2-678 FUSETOOLS2-1172 Fix start integration with resource and provide test

apupier:
startIntegrationWithResource

### DIFF
--- a/src/IntegrationUtils.ts
+++ b/src/IntegrationUtils.ts
@@ -34,13 +34,13 @@ const devModeIntegration: string = 'Dev Mode - Apache Camel K Integration in Dev
 export const basicIntegration: string = 'Basic - Apache Camel K Integration without extra options';
 export const configMapIntegration: string = 'ConfigMap - Apache Camel K Integration with Kubernetes ConfigMap';
 export const secretIntegration: string = 'Secret - Apache Camel K Integration with Kubernetes Secret';
-const resourceIntegration: string = 'Resource - Apache Camel K Integration with Resource file';
+export const resourceIntegration: string = 'Resource - Apache Camel K Integration with Resource file';
 export const propertyIntegration: string = 'Property - Apache Camel K Integration with Property';
 const dependencyIntegration: string = 'Dependencies - Apache Camel K Integration with Explicit Dependencies';
 export const vscodeTasksIntegration: string = 'Use a predefined Task - useful for multi-attributes deployment';
 
 const ResourceOptions: vscode.OpenDialogOptions = {
-	canSelectMany: true,
+	canSelectMany: false,
 	openLabel: 'Open Resource File(s)',
 	filters: {
 		'Text files': ['txt'],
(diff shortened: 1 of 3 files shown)


src/IntegrationUtils.ts
```diff
@@ -464,12 +457,7 @@ export function computeKamelArgs(absoluteRoot: string,
 		kamelArgs.push(`--profile=${profile}`);
 	}
 	if (resource && resource.trim().length > 0) {
-		let resourceArray = resource.split(' ');
-		if (resourceArray && resourceArray.length > 0) {
-			resourceArray.forEach(res => {
-				kamelArgs.push(`--resource="${res}"`);
-			});
-		}
+		kamelArgs.push(`--resource=${resource}`);
```
apupier, line 460: 2 things here:
* remove double-quotes which are nto working. The spaces are handled correctly by using different parameters passed to the child_process
* remove the support of the multiple resources. Space character was used a seprator inside a string. it doesn't work when there is a space in the path of one of the resource. Given the previous problem of double-quotes, it has never worked anyway, so better removing for now to simplify code and introduce in another PR the support of multiple resources using a data structure instead of merging/splitting strings.

src/IntegrationUtils.ts
```diff
@@ -34,13 +34,13 @@ const devModeIntegration: string = 'Dev Mode - Apache Camel K Integration in Dev
...
 export const secretIntegration: string = 'Secret - Apache Camel K Integration with Kubernetes Secret';
-const resourceIntegration: string = 'Resource - Apache Camel K Integration with Resource file';
+export const resourceIntegration: string = 'Resource - Apache Camel K Integration with Resource file';
 const propertyIntegration: string = 'Property - Apache Camel K Integration with Property';
 const dependencyIntegration: string = 'Dependencies - Apache Camel K Integration with Explicit Dependencies';
 export const vscodeTasksIntegration: string = 'Use a predefined Task - useful for multi-attributes deployment';
 
 const ResourceOptions: vscode.OpenDialogOptions = {
-	canSelectMany: true,
+	canSelectMany: false,
```
apupier, line 43: this is the place limiting resource selection to a single one

src/IntegrationUtils.ts
```diff
@@ -289,20 +289,13 @@ function getSelectedSecret(): Promise<string | undefined> {
...
 function getSelectedResource(): Promise<string> {
 	return new Promise<string> ( async (resolve, reject) => {
-		let returnedResources: string = '';
-		const fileUri = await vscode.window.showOpenDialog(ResourceOptions);
-		if (fileUri === undefined || fileUri.length === 0) {
-			reject(new Error("No Resource file(s) specified."));
+		const fileUris = await vscode.window.showOpenDialog(ResourceOptions);
+		if (fileUris === undefined || fileUris.length === 0) {
+			reject(new Error('No Resource file(s) specified.'));
+		} else if(fileUris.length >= 2) {
```
lhein, line 295: Why do you use `>=2` instead of `>1`?

lhein (approved):

src/IntegrationUtils.ts
apupier: I found it clearer that the case of `1` was not part of any of the written conditions by not writing it.

src/IntegrationUtils.ts
bfitzpat: Don't we need the double quotes for any resources with spaces in the path?

apupier: the test https://github.com/camel-tooling/vscode-camelk/pull/823/files#diff-b7ed3e6f1115050b03813503cf26ba0ef985fc1ca54ef08e59f581d50c55f9e6R78 is proving that we do not need them (at least on Linux as we do not have Windows CI with kamel instance). In fact, it was the root cause of non-working `Start integration with resource` command.

Based on https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback, the double-quotes would be needed only for program path.

```
exec('"/path/to/test file/test.sh" arg1 arg2');
// Double quotes are used so that the space in the path is not interpreted as
// a delimiter of multiple arguments.
```

For the arguments, it is not mentioned that we need to escape any characters.

bfitzpat (approved):

apupier:
build is green https://app.circleci.com/pipelines/github/camel-tooling/vscode-camelk/1038/workflows/4ac84992-92fb-433d-b363-78a2c3d4c770/jobs/1327 even if not updated in github UI

PR #835, opened 2021-07-08
DRAFT with just an initialize method

apupier:
the telemetry attribute retrieved by the stub is undefined:

https://app.circleci.com/pipelines/github/camel-tooling/vscode-camelk/1065/workflows/107d9edc-abd8-480a-9476-5677ef87c741/jobs/1355?invite=true#step-110-442

### DIFF
--- a/package.json
+++ b/package.json
@@ -368,7 +368,7 @@
 		"redhat.vscode-yaml"
 	],
 	"dependencies": {
-		"@redhat-developer/vscode-redhat-telemetry": "^0.1.1",
+		"@redhat-developer/vscode-redhat-telemetry": "^0.2.0",
 		"@types/detect-port": "^1.3.1",
 		"@types/request-promise": "^4.1.47",
 		"@types/shelljs": "^0.8.9",
--- a/package.json
+++ b/package.json
@@ -365,8 +365,7 @@
 	},
 	"extensionDependencies": [
 		"ms-kubernetes-tools.vscode-kubernetes-tools",
-		"redhat.vscode-yaml",
-		"redhat.vscode-commons"
+		"redhat.vscode-yaml"
 	],
 	"dependencies": {
 		"@redhat-developer/vscode-redhat-telemetry": "^0.1.1",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -65,6 +65,7 @@ export const COMMAND_ID_START_INTEGRATION = 'camelk.startintegration';
 const COMMAND_ID_OPEN_OPERATOR_LOG = 'camelk.integrations.openOperatorLog';
 export const COMMAND_ID_START_JAVA_DEBUG = 'camelk.integrations.debug.java';
 export async function activate(context: vscode.ExtensionContext) {
+	console.log('Activate VS Code Camel K');
 	stashedContext = context;
 	await telemetry.initializeTelemetry(context);
 	camelKIntegrationsProvider = new CamelKNodeProvider(context);
@@ -213,7 +214,7 @@ export async function activate(context: vscode.ExtensionContext) {
 	
 	return {
 		getStashedContext() : vscode.ExtensionContext {
-			return stashedContext;
+			return stashedContext;Will initialize Telemetry
 		},
 		getCamelKIntegrationsProvider(): CamelKNodeProvider {
 			return camelKIntegrationsProvider;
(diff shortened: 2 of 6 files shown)


fbricon:
That looks good on paper. Now the test failure seems to indicate the service is not initialized during testing

src/extension.ts
```diff
@@ -66,6 +66,7 @@ const COMMAND_ID_OPEN_OPERATOR_LOG = 'camelk.integrations.openOperatorLog';
 export const COMMAND_ID_START_JAVA_DEBUG = 'camelk.integrations.debug.java';
 export async function activate(context: vscode.ExtensionContext) {
 	stashedContext = context;
+	telemetry.initializeTelemetry(context);
```
fbricon, line 69: try to await here?

apupier: still same error I do not think that it is not a problem of timing. When I tried initially, i put traces and they were written in correct order.

apupier:
> That looks good on paper.

agree. i do not understand why it doesn't work. it sounds like there are different instances although it is a kind of static access

> Now the test failure seems to indicate the service is not initialized during testing

When initially tried, i put traces and it was showing that the initialization was done but the telemetry retrieved by the mock framework is not the one used by the product during test execution.

package.json
```diff
@@ -365,8 +365,7 @@
 	},
 	"extensionDependencies": [
 		"ms-kubernetes-tools.vscode-kubernetes-tools",
-		"redhat.vscode-yaml",
-		"redhat.vscode-commons"
+		"redhat.vscode-yaml"
 	],
 	"dependencies": {
 		"@redhat-developer/vscode-redhat-telemetry": "^0.1.1",
```
fbricon, line 371: ^0.2.0

apupier: I was trying to keep the change minimal to demonstrate it. But updated and same error https://app.circleci.com/pipelines/github/camel-tooling/vscode-camelk/1069/workflows/06ba6b81-9666-487c-8bf1-3c3997b9b51e/jobs/1359?invite=true#step-110-452

fbricon:
So I can see Telemetry.ts is being instanciated twice, once when the extensions' activate function is called, and another time. That second time, the initialize method is not called, explaining the failures.
So most likely, 2 sets of tests are being run (UI vs other?). Whatever is launching the 2nd type of test should ensure initialize is being called.

src/extension.ts
```diff
@@ -212,7 +214,7 @@ export async function activate(context: vscode.ExtensionContext) {
 	
 	return {
 		getStashedContext() : vscode.ExtensionContext {
-			return stashedContext;
+			return stashedContext;Will initialize Telemetry
```
fbricon, line 217: is that a comment?

apupier: that's a bad copy when I accidently click on the mouse scroll wheel. i will remove it

apupier:
siwtched back to VS Code 1.57.1 to continue testing. And wouhou, the mock is better workign when the coverage is deactivated.

So now the question is how to have both of them workign at the same time? Maybe need to switch to nyc as istanbul is deprecated.
or we deactivate the coverage for now to have simpler code. The coverage is using deprecated api anyway.

fbricon:
Maybe call telemetry.initialize explicitly when starting the code coverage (use a mock context if needed)

apupier:
> Maybe call telemetry.initialize explicitly when starting the code coverage (use a mock context if needed)

The problem will be that I won't test anymore that the Telemetry is initialized when the activate method is called. Or need to add another tests checking for that with spies but it seems that we start to have weak and complex tests. Not sure it would be worthy.

I think I would prefer to deactivate coverage for some time and plan a migration to nyc and find a way to have it working properly.

src/test/runTests.ts
```diff
@@ -14,7 +14,8 @@ async function main() : Promise<void> {
 		if(vscodeVersionForTest !== undefined){
 			vscodeExecutablePath = await downloadAndUnzipVSCode(vscodeVersionForTest);
 		} else {
-			vscodeExecutablePath = await downloadAndUnzipVSCode('stable');
+			// FIXME: using previous version to workaround https://github.com/microsoft/vscode/issues/126636
+			vscodeExecutablePath = await downloadAndUnzipVSCode('1.57.1');
 		}
 		console.log(`vscodeExecutablePath = ${vscodeExecutablePath}`);
 		const testWorkspace = path.resolve(__dirname, '../../../test Fixture with speci@l chars');
```
fbricon, line 21: need to remove installExtraExtension(cliPath, 'redhat.vscode-commons'); down there

apupier:
created cleaner PRs:
- to remove istanbul: https://github.com/camel-tooling/vscode-camelk/pull/842
- use new API https://github.com/camel-tooling/vscode-camelk/pull/844
- upgrade telemetry package https://github.com/camel-tooling/vscode-camelk/pull/845
@fbricon thanks a lot for the help

PR #840, opened 2021-07-12
FUSETOOLS2-1200 - using previous VS Code version in test

apupier:
to workaround https://github.com/microsoft/vscode/issues/126636

Given that there is currently no workaround defined upstream, i think we should go with testing with previous version to unblock master

### DIFF
--- a/src/test/runTests.ts
+++ b/src/test/runTests.ts
@@ -14,7 +14,8 @@ async function main() : Promise<void> {
 		if(vscodeVersionForTest !== undefined){
 			vscodeExecutablePath = await downloadAndUnzipVSCode(vscodeVersionForTest);
 		} else {
-			vscodeExecutablePath = await downloadAndUnzipVSCode('stable');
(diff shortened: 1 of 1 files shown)


lhein (approved):

bfitzpat (approved):
Thinking we should probably do something similar with the rest of our extensions as well, at least until there's a fix upstream

apupier:
> Thinking we should probably do something similar with the rest of our extensions as well, at least until there's a fix upstream

yes for the ones that are using a folder as workspace when launchign tests

apupier:
> > Thinking we should probably do something similar with the rest of our extensions as well, at least until there's a fix upstream
> 
> yes for the ones that are using a folder as workspace when launchign tests

it should be even more rpecise, only the ones that are using a folder as workspace when launchign tests and that requires a trusted workspace.
Otherwise, a simple workaround consists in deactivating the Workspace Trust (there is an option for that)

PR #842, opened 2021-07-12
FUSETOOLS2-1202 - remove Istanbul coverage

apupier:
it is causing troubles with Mock framework and it is deprecated for
several years

there is https://issues.redhat.com/browse/FUSETOOLS2-226 to add coverage back


bfitzpat (approved):
Makes sense

PR #853, opened 2021-07-16
FUSETOOLS2-1208 - Provide CodeLens to refresh classpath dependencies

apupier:
refreshClasspathDependencies

Currently, the command to refresh classpath dependency is provided as a
command which requires the user to know that it exists. By providing it
as codelens in this PR on top of a Camel  K modeline inside Java files, it will
spread knowledge of this capability. Also it will require less click.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ Note: By default, `Auto-scroll` is checked and new entries in the log will autom
 
 To benefit from Java Language completion on standalone files, [VS Code Language support for java](https://marketplace.visualstudio.com/items?itemName=redhat.java) needs to be installed. An invisible project is created with a default classpath.
 
-The command `Refresh local Java classpath for Camel K standalone file based on current editor. Available with kamel 1.4+.` allows to have specfic dependencies of the opened Integration part of the classpath. For isntance, specific dependencies can be decalred as modeline.
+The command `Refresh local Java classpath for Camel K standalone file based on current editor. Available with kamel 1.4+.` refreshes specific dependencies for the classpath of the opened Integration, such as dependencies declared as part of the modeline of the file. When using a modeline to configure such dependencies, you gain a [CodeLens](https://code.visualstudio.com/blogs/2017/02/12/code-lens-roundup) link (`Refresh classpath dependencies`) at the top of the editor to trigger the refresh more easily.
 
 Be aware of the following limitations:
   - It requires Camel K 1.4.0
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -265,7 +265,7 @@ Note: By default, `Auto-scroll` is checked and new entries in the log will autom
 
 To benefit from Java Language completion on standalone files, [VS Code Language support for java](https://marketplace.visualstudio.com/items?itemName=redhat.java) needs to be installed. An invisible project is created with a default classpath.
 
-The command `Refresh local Java classpath for Camel K standalone file based on current editor. Available with kamel 1.4+.` allows to have specfic dependencies of the opened Integration part of the classpath. For isntance, specific dependencies can be decalred as modeline.
+The command `Refresh local Java classpath for Camel K standalone file based on current editor. Available with kamel 1.4+.` allows to have specific dependencies of the opened Integration part of the classpath. For instance, specific dependencies can be declared as modeline. When a Camel K modeline is provided, a CodeLens is provided to refresh the classpath dependencies.
```
bfitzpat, line 268: This seems a bit awkward as a paragraph to me. I doubt that many users will know what a CodeLens is. It may be useful to provide a link to https://code.visualstudio.com/blogs/2017/02/12/code-lens-roundup or http://theblogreaders.com/visual-studio-code-tips-codelens-visual-studio-code/ or some other source (not a lot of info on CodeLens from a user perspective in the VS Code site beyond API references). 

Also I might suggest a few tweaks to the language here...

"The command X refreshes specific dependencies for the classpath of the opened Integration, such as dependencies declared as part of the modeline of the file. When using a modeline to configure such dependencies, you gain a CodeLens link ("Refresh classpath dependencies") at the top of the editor to trigger the refresh more easily."

src/codelenses/ClasspathRefreshCodeLensProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+	
+	onDidChangeCodeLenses?: vscode.Event<void>;
+	
+	provideCodeLenses(document: vscode.TextDocument): vscode.ProviderResult<vscode.CodeLens[]> {
+		const fulltext = document.getText();
+		if(fulltext.includes(CAMELK_MODELINE_PREFIX)) {
+			const topOfDocument = new vscode.Range(0, 0, 0, 0);
+			const classPathRefreshCommand: vscode.Command = {
+				command: extension.COMMAND_ID_CLASSPATH_REFRESH,
+				title: 'Refresh classpath dependencies',
```
bfitzpat, line 32: Might be good to externalize this as a constant for easier modification later?

bfitzpat (approved):
I really like this approach. It bubbles up an easier method for the user to refresh their classpath rather than trying to hunt for the right command in the palette. I just have a couple of adjustments to suggest around the language itself, but they're more for clarification purposes than anything else.

bfitzpat (approved):

PR #854, opened 2021-07-19
FUSETOOLS2-1108 - escape space in path to bin path

apupier:
* it was done when using an exec and concatenating arguments but not
when using spawn.
* not adding test as it is tested with the Insiders job (for which the
Code instance is deployed inside a "Code - Insiders" instead of simply
"Code" folder and so containing space).


apupier:
maybe needed only on Windows https://nodejs.org/api/child_process.html#child_process_spawning_bat_and_cmd_files_on_windows and so Inisders would fail for another reason

PR #881, opened 2021-08-06
Upgrade base image on Circle CI to ubuntu-2004:202107-02

apupier:


apupier:
sound slike we woudl need to do code change on package-lock.json:

```
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
```
(and then failing)

apupier:
migrating to GitHub Actions so no more relevant

PR #997, opened 2022-01-10
FUSETOOLS2-1340 - detect potential leak of dev dependencies

apupier:
- tslint was not well configured, so instead of configuring the old and
deprecated tslint, I moved to eslint trying to stick with VS Code
Didact.
- fixed error reported issues except `no-async-promise-executor` which
is deactivated as I do not know how to fix it safely. Reported
https://issues.redhat.com/browse/FUSETOOLS2-1415
- notably mkdirp was not explicitly declared


lhein (approved):
quite a bit more in this PR than just leak detection ;)

PR #770, opened 2021-05-18
#769: Use an enum to detected the OS platform

nanomad:
Use an enum to detected the OS platform

    Then decode it differently for kamel and kubectl.

    Fixes #769

Issue #769, opened 2021-05-18
kamel cli download not working on windows


nanomad:
Extension version: 0.0.25

On windows, `getDownloadURLForCamelKTag` uses win32 as the platform name, but download links contain `windows-64bit` instead.

Since version checking is mandatory, the extension cannot start.

apupier:
Thanks for the report

nota: I created https://issues.redhat.com/browse/FUSETOOLS2-1140 for Red Hat team to look at it

nanomad:
Looking at the code, the same applies to camel-k on MacOS: platform name is `darwin`, but download links have `mac`

apupier:
the constants https://github.com/camel-tooling/vscode-camelk/blob/6abf3fec07ec7a07c2c773de57b8cc6450ebc314/src/shell.ts#L22-L24 should be different for `kamel` and `kubectl`. Currently, the same are used.

nanomad:
Considering that the kubectl does some "renaming" for windows already (win32 -> windows) I think the best approach could be:
- Use an enum in platform detection code (shell.ts)
- Have `kubectl` translate them one way
- Have `kamel` translate them another way

If that's OK I can provide a quick PR

apupier:
> If that's OK I can provide a quick PR

of course, this is very welcome

nanomad:
PR Submitted: https://github.com/camel-tooling/vscode-camelk/pull/770

PR #1150, opened 2022-06-08
[FUSETOOLS2-1140] Fix enum for windows platform

unsortedhashsets:
Enum value for the windows platform was `0`

As a result of this value [condition](https://github.com/camel-tooling/vscode-camelk/blob/main/src/installer.ts#L137) received 0(false) in it.

And consequently, the stage of assembling the URL for the subsequent installation of the `kamel` was skipped with the error:

`Camel K CLI Version 1.9.2 unavailable. Please check the Apache Camel K version specified in VS Code Settings. Inaccessible url:`

More information:
[1] Initial issue - https://github.com/camel-tooling/vscode-camelk/issues/769
[2] RH issue - https://issues.redhat.com/browse/FUSETOOLS2-1140
[3] PR with enum addition - https://github.com/camel-tooling/vscode-camelk/pull/770

### DIFF
--- a/src/shell.ts
+++ b/src/shell.ts
@@ -20,18 +20,18 @@ import { ChildProcess } from 'child_process';
 import * as shelljs from 'shelljs';
 
 const WINDOWS = 'win32';
-const MACOS  = 'darwin';
-const LINUX  = 'linux';
+const MACOS = 'darwin';
+const LINUX = 'linux';
 
 export enum Platform {
-    WINDOWS,
+    WINDOWS = 1,
     MACOS,
     LINUX
 }
 
 export interface FindBinaryResult {
-	err: number | null;
-	output: string;
+    err: number | null;
+    output: string;
 }
 
 export function concatIfBoth(s1: string | undefined, s2: string | undefined): string | undefined {
@@ -57,7 +57,7 @@ export function isUnix(): boolean {
     return (process.platform === LINUX);
 }
 
-export function getPlatform() : Platform | undefined {
+export function getPlatform(): Platform | undefined {
     if (isWindows()) { return Platform.WINDOWS; }
     if (isMacOS()) { return Platform.MACOS; }
     if (isUnix()) { return Platform.LINUX; }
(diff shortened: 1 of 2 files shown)


apupier:
> Enum value for the windows platform was `0`
> 
> As a result of this value [condition](https://github.com/camel-tooling/vscode-camelk/blob/main/src/installer.ts#L137) received 0(false) in it.

OMG, good catch. What a strange language than TypeScript/JavaScript

apupier (commented):
It would be nice if you can add the issue number in the commit message. It allows to have more context in the git history to understand some specific changes.

src/shell.ts
```diff
@@ -20,13 +20,13 @@ import { ChildProcess } from 'child_process';
...
-const MACOS  = 'darwin';
-const LINUX  = 'linux';
+const MACOS = 'darwin';
+const LINUX = 'linux';
 
 export enum Platform {
-    WINDOWS,
-    MACOS,
-    LINUX
+    WINDOWS = 1,
```
joshiraez, line 27: Wondering if it would be nice to add NONE or INVALID with string '' and num value 0.

Pros:
Clearly shows the intent that number 0 shouldn't be used and is an invalid value
Would fail for the cases where it should be failing (when checking if platform exists, undefined, null, and the NONE value (either empty string or 0 depending on using the const or the enum value) would all return false.

Cons:
False positives when checking for undefined variables.

Thoughts?

src/shell.ts
```diff
@@ -20,13 +20,13 @@ import { ChildProcess } from 'child_process';
...
+    MACOS = 2,
+    LINUX = 3
```
apupier, line 29: based on https://www.tutorialsteacher.com/typescript/typescript-enum , we can set only the first numeric enum value.
so would be:
```suggestion
    WINDOWS = 1,
    MACOS,
    LINUX
```

apupier (approved):

unsortedhashsets:
Rebased to launch circle ci

unsortedhashsets:
Rebased with main / commits squash / rename commit and pr

PR #1155, opened 2022-06-14
FUSETOOLS2-1645: Camel K extension cannot be compiled on WIN because ESLint

djelinek:
Signed-off-by: Dominik Jelinek <djelinek@redhat.com>


djelinek:
Hi @apupier any idea why the PR check pipeline was not executed and is still in status 'pending'?

apupier:
I guess there was a GitHub outage; Can you try to push again (with a very small modification in commit message for instance)?

apupier (approved):
Sounds strange that it is required to use it this way on Windows. Can it depends on the way eslint/npm is installed?

djelinek:
> Sounds strange that it is required to use it this way on Windows. Can it depends on the way eslint/npm is installed?

I am not sure, I was not digging that much into deep. It could be related to how exactly is eslint installed - globally or at extension level

Main reason is I was struggling on windows because I hit error with `unknown '.'` when build was trying to execute `./node_modules/.bin/eslint src --ext .ts` ... after few attempts I realised that in powershell the command starting with '.' causing this issue

Calling eslint directly was working for me on all platforms

PR #1156, opened 2022-06-15
FUSETOOLS-1618: [vscode-camelk] Stabilize UI tests pipelines on all operating systems

djelinek:
Signed-off-by: Dominik Jelinek <djelinek@redhat.com>

### DIFF
--- a/src/ui-test/camelk_extension_test.ts
+++ b/src/ui-test/camelk_extension_test.ts
@@ -1,35 +1,49 @@
 import * as pjson from '../../package.json';
 import { assert } from 'chai';
+import { EditorView, ExtensionsViewItem, VSBrowser, WebDriver } from 'vscode-extension-tester';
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -1,8 +1,8361 @@
 {
 	"name": "vscode-camelk",
 	"version": "0.0.32",
-	"lockfileVersion": 1,
+	"lockfileVersion": 2,
```
apupier, line 4: woudl be better to upgrade the lock file version in a separated Pull Request

apupier (changes requested):
There are no changes in dependencies so the package-lock.json should not be modified. I guess you called a `npm install` locally instead of a `npm ci`

package-lock.json
djelinek: You are right, I've removed unnecessary package-lock.json changes from this PR

PR #1186, opened 2022-07-12
[FUSETOOLS2-1653] Upgrade strip-ansi to 7.0.1

joshiraez:
Fixes https://issues.redhat.com/browse/FUSETOOLS2-1653

Resources: 
* https://github.com/aws/aws-toolkit-vscode/issues/2285
* https://github.com/aws/aws-toolkit-vscode/pull/1775
* https://www.typescriptlang.org/tsconfig#esModuleInterop

Strip-ansi 7.0 is a pure ES module so there is a need to enable esModuleInterop. This causes that we need to change some imports on those modules that are not ES6 modules (those whose files don't have the export ___ in the end) to not use the `import * as ____ from ____` and instead put those imports as `import ____ from ____`

This PR adds the esModuleInterop compiler version, fixes the imports that break after it, and bumps the version in strip-ansi


package-lock.json
```diff
@@ -1,8 +1,8249 @@
 {
 	"name": "vscode-camelk",
 	"version": "0.0.32",
-	"lockfileVersion": 1,
+	"lockfileVersion": 2,
```
apupier, line 4: the package-lock.json wasn't upgraded yet to a recent npm versiobn. it would have been better to do the upgrade in another PR as it will allow to review the changes. Fine for this time.

package.json
```diff
@@ -399,10 +399,16 @@
...
 		"tar": "^6.1.11",
 		"tmp": "^0.2.1",
 		"valid-filename": "^3.1.0",
 		"vscode-kubernetes-tools-api": "^1.3.0"
+	},
+	"__metadata": {
+		"id": "5181e15d-d18d-444e-ba3b-e0cf48cab6f8",
+		"publisherDisplayName": "Red Hat",
+		"publisherId": "eed56242-9699-4317-8bc7-e9f4b9bdd3ff",
+		"isPreReleaseVersion": false
```
apupier, line 412: why is it added?

test Fixture with speci@l chars/TestDeployInSpecificNamespace.java
```diff
@@ -0,0 +1,17 @@
...
+
+      // Write your routes here, for example:
+      from("timer:java?period=1000")
+        .routeId("java")
+        .setBody()
+          .simple("Hello Camel K from ${routeId}")
+        .to("log:info");
+
+  }
+}
```
apupier, line 17: I think this file is unrealted to this change

apupier (changes requested):
good findings with the esModuleInterOp

test Fixture with speci@l chars/TestDeployInSpecificNamespace.java
joshiraez: Wops my bad. I'll remove it

package.json
joshiraez: I'm not sure. I'll remove it and try if it breaks anything. It might be some leftover of some node script.

apupier (approved):

PR #1003, opened 2022-01-13
WIP FUSETOOLS2-1126 - provide Contextual Java debug only on running

apupier:
extension

- test to implement

current error:
```
  1) Tooling for Apache Camel K extension
       Java Debug
         Create File Deploy it and Check Java Debug available:
     ElementNotVisibleError: element not interactable
  (Session info: chrome=91.0.4472.164)
  (Driver info: chromedriver=91.0.4472.101 (af52a90bf87030dd1523486a1cd3ae25c5d76c9b-refs/branch-heads/4472@{#1462}),platform=Linux 5.15.13-200.fc35.x86_64 x86_64)
      at Object.checkLegacyResponse (node_modules/selenium-webdriver/lib/error.js:546:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:509:13)
      at /home/apupier/git/vscode-camelk/node_modules/selenium-webdriver/lib/http.js:441:30
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
  From: Task: WebElement.click()
      at Driver.schedule (node_modules/selenium-webdriver/lib/webdriver.js:807:17)
      at WebElement.schedule_ (node_modules/selenium-webdriver/lib/webdriver.js:2010:25)
      at WebElement.click (node_modules/selenium-webdriver/lib/webdriver.js:2092:17)
      at InputBox.<anonymous> (node_modules/monaco-page-objects/out/components/workbench/input/Input.js:74:25)
      at Generator.next (<anonymous>)
      at fulfilled (node_modules/monaco-page-objects/out/components/workbench/input/Input.js:5:58)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)

```

PR #1190, opened 2022-07-18
[FUSETOOLS2-941] Tests for entering java debug on camel k

joshiraez:
Original PR in https://github.com/camel-tooling/vscode-camelk/pull/1003

Attempted to stabilize aurelien existing test and add another one that asserts that we enter debug mode.

Both aurelien tests "check if java debug is available in valid files" and "check java debug not available on bad camel files" work well now.

The one that checks if we can enter debug mode it's skipped as there is currently a problem where when we enter the java debug mode on the camel k integration, an error pops out saying "Java Debug is not configured" on VSCode test instance. I tried installing the java debugger through Extester (wasn't able to, even though the log says it's installed and if you input a bad ID the test aborts with an error, so, it's working, somehow?) or sleeping the test instance, installing it manually and letting them continue normally. The code "should" work though, except that the method "toggleBreakpoint" doesn't seem to work either (it doesn't do anythng aparently when watching the VSCode test instance). Maybe someone from the QE team could take a look?


src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+async function getContextMenuOf(integrationLabel: string) {
+	const item = await findIntegrationOnSideBar(integrationLabel);
+	return await item.openContextMenu()
+}
+
+async function startDebuggerOn(integrationLabel: string) {
+	const menu = await getContextMenuOf(integrationLabel);
+
+	await menu.getItem(START_DEBUG_LABEL).then(item => item?.click());
```
apupier, line 195: I would recommend to not mix the 2 notations of `await` and `then`.
Having await everywhere allows to be more consistent.

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+async function findIntegrationOnSideBar(integrationLabel: string) {
+	const section = await getIntegrationSectionFromSideView(integrationLabel);
+	return await section.findItem(integrationLabel) as ViewItem;
+}
+
+async function getIntegrationSectionFromSideView(integrationLabel: string) {
+	const section = await new SideBarView().getContent().getSection('Apache Camel K Integrations') as CustomTreeSection;
+	await section.expand();
+	await VSBrowser.instance.driver.sleep(5000);
```
apupier, line 169: sleep should be avoided as much as possible. it is introducing slowness and flakiness to test suite. it will be better to use a waitUntil with a specific condition

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+});
+async function startIntegrationOnCurrentFile() {
+	const workbench = new Workbench();
+	await workbench.executeCommand('Start Apache Camel K Integration');
+	console.log('Start command');
+	const startMode = await InputBox.create();
+	await startMode.selectQuickPick('Basic');
+
+	await VSBrowser.instance.driver.sleep(1000);
```
apupier, line 109: sleep should be avoided as much as possible. it is introducing slowness and flakiness to test suite. it will be better to use a waitUntil with a specific condition

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,201 @@
...
+			assert.isTrue(await menu.hasItem(START_DEBUG_LABEL));
+		});
+
+		it.skip('Check Java Debug stops at breakpoint', async function() {
+			//Currently failing because "No Java Debugger configured" error window
+			//It won't work even if I install the extension using ExTester or manually on the VSCode Instance
+			this.timeout(30000);
+			await addBreakpointToFile();
+			await startDebuggerOn(INTEGRATION_LABEL);
+			await VSBrowser.instance.driver.sleep(3000);
```
apupier, line 59: sleep should be avoided as much as possible. it is introducing slowness and flakiness to test suite. it will be better to use a waitUntil with a specific condition

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,201 @@
...
+	describe('Java Debug', function () {
+		
+		const INTEGRATION_LABEL = 'java-debug-test';
+		const INTEGRATION_FILE = 'JavaDebugTest';
+
+		before(async function (){
+			this.timeout(20000);
+			await createIntegration(INTEGRATION_FILE);
+			await startIntegrationOnCurrentFile();
+			await VSBrowser.instance.driver.sleep(5000);
```
apupier, line 41: sleep should be avoided as much as possible. it is introducing slowness and flakiness to test suite. it will be better to use a waitUntil with a specific condition

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,201 @@
...
+const START_DEBUG_LABEL = 'Start Java debugger on Camel K integration';
+const REMOVE_INTEGRATION_LABEL = 'Remove Apache Camel K Integration';
+
+
+describe('Tooling for Apache Camel K extension', function () {
+
+	before(async function () {
+		this.timeout(200000);
+		await prepareTempWorkspaceForTests(WORKSPACE_FOLDER);
+		await VSBrowser.instance.driver.sleep(3000);
```
apupier, line 29: sleep should be avoided as much as possible. it is introducing slowness and flakiness to test suite. it will be better to use a waitUntil with a specific condition

apupier (changes requested):
the integration deploye din the test is not cleaned at the end of the test. i think it is the reason of all the failures on the Ci https://app.circleci.com/pipelines/github/camel-tooling/vscode-camelk/2446/workflows/4262c410-0f59-4f79-b480-9b8134280492/jobs/2814

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,201 @@
...
+			this.timeout(20000);
+
+			const item = await findIntegrationOnSideBar(INTEGRATION_LABEL);
+			const menu = await item.openContextMenu();
+
+			assert.isTrue(await menu.hasItem(START_DEBUG_LABEL));
+		});
+
+		it.skip('Check Java Debug stops at breakpoint', async function() {
+			//Currently failing because "No Java Debugger configured" error window
```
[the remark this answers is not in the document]
joshiraez: I tried just that after Dominik told me but it didn't appear to be working (the install from marketplace method). If I got into the instance I could see the extension wasn't installed.

In any case, even installing the extension didn't solve the configuration problem (tried installing it manually in the vscode test instance while it was slept, and then leaving the window at where the tests expected after making sure the extension was installed, but there was still the same error popping of "no Java Debug configured")

src/ui-test/camelk_java_debug_test.ts
joshiraez: I can try using waitUntil but I'm not sure what conditions do we have to know when something has finished on the ui-tests. None of the other ui-tests use it either, only the integration tests. I'll review it and see what I can do.

joshiraez:
The CI tests passed after retrying once. There is some flaky tests on the other test suites I think? The tests failing were in the middle of the new dev UI tests so I'm not sure what might be happening. I did several tries on local and the tests seemed robust when ran along with the others.

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+async function startDebuggerOn(integrationLabel: string) {
+	const menu = await getContextMenuOf(integrationLabel);
+
+	await menu.getItem(START_DEBUG_LABEL).then(item => item?.click());
+}
+
+async function isThereSomethingInVariablesInDebugger() {
+	const section = await new SideBarView().getContent().getSection('Variables') as CustomTreeSection;
+	return await (await section.getVisibleItems()).length > 0;
```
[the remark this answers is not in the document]
joshiraez: You are right the leftmost await is not needed, but I need the first one to resolve the promise.

I don't think waitUntil is what I want here, as we don't know if the condition might even happen (if the test fails)

src/ui-test/camelk_java_debug_test.ts
apupier: My bad, Selenium is coming with its own way to provide a conditional wait using driver.wait, see for instance https://github.com/camel-tooling/vscode-camelk/blob/06972bb32366d2767c1bb2b052c5b6396ce9ee19/src/ui-test/dev_integration_test.ts#L89 (instead of waitUntil)

On what to wait, I do not know, it depends what is expected by prepareTempWorkspaceForTests, I would even say that the conditional wait shoudl be inside this method. If not possible, the driver.wait must be related to the next first action.

src/ui-test/camelk_java_debug_test.ts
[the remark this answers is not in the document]
apupier: I do not understand. Do yuou mean that manually you were not able to have the feature working at all?

src/ui-test/camelk_java_debug_test.ts
joshiraez: I managed to change every one using the driver.wait :tada:

joshiraez:
After checking this with @djelinek we came to the conclusion that the test break on the latest vscode extension, just like in AtlasMap https://github.com/redhat-developer/vscode-extension-tester/issues/466

I'm going to pause development in this for the moment and move to other feats in the epic while we wait for upstream to fix these issues

src/ui-test/camelk_java_debug_test.ts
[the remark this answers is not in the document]
djelinek: Could you please try to different setup for uitests_runner? It looks like `setupAndRun` will override installed all vscode instance setup which was done till that time.

e.g. something similar this one:
```js
const tester = new ExTester(storageFolder, releaseType, extensionFolder);
await tester.downloadCode(process.env.CODE_VERSION);
await tester.downloadChromeDriver(process.env.CODE_VERSION);
await tester.installFromMarketplace('redhat.java');
await tester.installFromMarketplace('vscjava.vscode-java-debug');
await tester.runTests('out/src/ui-test/*extension_test.js', { vscodeVersion: process.env.CODE_VERSION });
```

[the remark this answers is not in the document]
joshiraez: I'll try this on Monday, at least manually to see the extension is actually there in the marketplace, but I dont think Ill implement this in the tests themselves until we are unblocked.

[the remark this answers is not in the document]
apupier: there was missing the Camel K extension with code in previous comment, I think it will be more somethign like:

```
	const tester = new ExTester(storageFolder, releaseType, extensionFolder);
	await tester.downloadCode(process.env.CODE_VERSION);
	await tester.downloadChromeDriver(process.env.CODE_VERSION);
	await tester.installFromMarketplace('redhat.java');
	await tester.installFromMarketplace('vscjava.vscode-java-debug');
	await tester.installVsix();
	await tester.runTests('out/src/ui-test/*java_debug_test.js', { vscodeVersion: process.env.CODE_VERSION });
```
	
notice the `await tester.installVsix();`

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,245 @@
...
+			await startIntegration(INTEGRATION_LABEL);
+		})
+
+		it('Check Java Debug available', async function () {
+			this.timeout(20000);
+
+			const item = await getIntegration(INTEGRATION_LABEL);
+			const menu = await item.openContextMenu();
+
+			assert.isTrue(await menu.hasItem(START_DEBUG_LABEL));
```
apupier, line 75: This needs a conditional waiting because the context menu becomes available only when the integration is in state "running". Which can take several seconds/minute it this test is played first

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,245 @@
...
+		});
+
+		it.skip('Check Java Debug stops at breakpoint', async function() {
+			//Currently failing because "No Java Debugger configured" error window
+			//It won't work even if I install the extension using ExTester or manually on the VSCode Instance
+			this.timeout(30000);
+			await addBreakpointToCurrentFile();
+			await startDebug(INTEGRATION_LABEL);
+			assert.isTrue(isThereSomethingInVariablesInDebugger());
```
apupier, line 84: this will always be false because we are looking if a Promise is true.
```suggestion
			assert.isTrue(await isThereSomethingInVariablesInDebugger());
```

apupier, line 84: also it needs to be a conditional wait. The variables in the debugger can take some time to be available as the integration is restarted and then Java debugger is attached

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,245 @@
...
+		it('Check Java Debug available', async function () {
+			this.timeout(20000);
+
+			const item = await getIntegration(INTEGRATION_LABEL);
+			const menu = await item.openContextMenu();
+
+			assert.isTrue(await menu.hasItem(START_DEBUG_LABEL));
+		});
+
+		it.skip('Check Java Debug stops at breakpoint', async function() {
```
apupier, line 78: Given all the problems detected with this test, i think it is better to split the Pull request and start by providing the tests only for the right-click menu available or not depending on the state of the integration.
And so get back to the limited scope of the initial work in this other PR and the FUSETOOLS2-1126
it will allow to finish the new feature and incorporate it in the main branch so that end-uers can benefit from it.

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,245 @@
...
+	const textEditor : TextEditor = new TextEditor();
+	await textEditor.toggleBreakpoint(13);
+	await textEditor.save();
+}
+
+async function startDebug(integrationLabel: string) {
+	const integration = await getIntegration(integrationLabel);
+	const menu = await integration.openContextMenu();
+	const item = await menu.getItem(START_DEBUG_LABEL)
+	item?.click();
```
apupier, line 224: requires an await

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,245 @@
...
+	VSBrowser.instance.driver.wait(isInDebugView);
```
apupier, line 225: requires an await

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,245 @@
...
+}
+
+async function isInDebugView() {
+	const control = await new ActivityBar().getViewControl('Run and Debug');
+	return control?.isSelected();
+}
+
+async function isThereSomethingInVariablesInDebugger() {
+	const section = await new SideBarView().getContent().getSection('Variables') as CustomTreeSection;
+	return (await section.getVisibleItems()).length > 0;
```
apupier, line 235: I think ti will need a conditional wait, something like:

```
	await VSBrowser.instance.driver.wait(async() => {
		const section = await new SideBarView().getContent().getSection('Variables') as CustomTreeSection;
		return (await section.getVisibleItems()).length > 0;
	}, 20000);
```

apupier (changes requested):

src/ui-test/camelk_java_debug_test.ts
joshiraez: I agree. I can remove the test and create a separate issue to do it at a later time.

joshiraez:
I've removed the test case for the debug session for now. I'll open an issue attached to the original issue talking about the missing test.

I also tried @djelinek (and @apupier ) code to install the Java Debugger but they don't seem to be working (screenshots from the test instance).
Screenshot from 2022-07-26 10-54-11
Screenshot from 2022-07-26 10-54-34
Screenshot from 2022-07-26 10-56-55

Another note is that I'm working https://issues.redhat.com/browse/FUSETOOLS2-1131 and it requires this as a starting branch. I have already started work on it and I'll be rebasing any new things we do to this branch

apupier:
> I also tried @djelinek (and @apupier ) code to install the Java Debugger but they don't seem to be working (screenshots from the test instance).

Can you show the exact code used and where please?

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,199 @@
...
+		await VSBrowser.instance.driver.wait(camelKToolingIsEnabled);
+	});
+
+	describe('Java Debug', function () {
+		
+		const INTEGRATION_LABEL = 'java-debug-test';
+		const INTEGRATION_FILE = 'JavaDebugTest';
+
+		before(async function (){
+			this.timeout(20000);
```
unsortedhashsets, line 61: Hi @joshiraez, I recommend to increase the timeout from 20000 to 60000.
I am not confident that tasks: 
- create integration
- start integration 
with/without `hasItem` check will be always so fast.

Example:
12:09:46    Test Debug on Camel K Integrations from Side Bar
12:10:05      Java Debug
12:10:27        1) "before all" hook for "Check Java Debug available"
12:10:28      No Java Debug on Invalid Files
12:10:50        2) "before all" hook for "Test Java Debugger Not Available On Invalid File"

12:16:58    1) Test Debug on Camel K Integrations from Side Bar
12:16:58         Java Debug
12:16:58           "before all" hook for "Check Java Debug available":
12:16:58       Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/mnt/hudson_workspace/workspace/vscode/vscode-camelk-pipeline/vscode-camelk/out/src/ui-test/camelk_java_debug_test.js)
12:16:58        at listOnTimeout (node:internal/timers:559:17)
12:16:58        at processTimers (node:internal/timers:502:7)
12:16:58  
12:16:58    2) Test Debug on Camel K Integrations from Side Bar
12:16:58         No Java Debug on Invalid Files
12:16:58           "before all" hook for "Test Java Debugger Not Available On Invalid File":
12:16:58       Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/mnt/hudson_workspace/workspace/vscode/vscode-camelk-pipeline/vscode-camelk/out/src/ui-test/camelk_java_debug_test.js)
12:16:58        at listOnTimeout (node:internal/timers:559:17)
12:16:58        at processTimers (node:internal/timers:502:7)

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,199 @@
...
+
+	});
+
+	describe('No Java Debug on Invalid Files', function() {
+
+		const INTEGRATION_LABEL = 'java-debug-test-invalid';
+		const INTEGRATION_FILE = 'JavaDebugTestInvalid';
+
+		before(async function (){
+			this.timeout(20000);
```
unsortedhashsets, line 90: Same as on line 61 (20000 -> 60000)

src/ui-test/camelk_java_debug_test.ts
apupier: I think this must be adressed

apupier (changes requested):
I think the commit history must be cleaned, here we have a lot of commits which are finally "reverted" which is making the history confusing.

You can also squash with my old "WIP" commit

joshiraez:
> > I also tried @djelinek (and @apupier ) code to install the Java Debugger but they don't seem to be working (screenshots from the test instance).
> 
> Can you show the exact code used and where please? Edit: o rmaybe in the oteh rPR as I think we do not need these extra extensions in this one.

I used the snippet you pasted exactly. I checked to see that those were the correct extensions IDs and they appear in the console.log

src/ui-test/camelk_java_debug_test.ts
joshiraez: Sorry, I didn't address this.

This is not needed because StartIntegration doesn't resolve until the integration is created/visible. Check:https://github.com/camel-tooling/vscode-camelk/pull/1190/files/8993fea59594abb190d6e85a7fdaaf4af18fee97#diff-da62a10e4200983f607ee8aeecd952e9941cdc88b1bc1c9ab1a93860ef672efbR198

apupier: I think it is waiting that the integration is visible but not that it is in "running" state. On first run, the "building" phase can take several minutes.

joshiraez: Oh, I see now. Ok, I'll put it

joshiraez:
@apupier I seem to have lost again the message after the squash.

The new commits address the conditional wait until the integration has started running. I added a few methods to assert on promises as it is what the wait method returns. I picked up the code from https://stackoverflow.com/a/65622470

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,198 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+import { assert } from 'chai';
```
apupier, line 20: src/ui-test/camelk_java_debug_test.ts:20:1 - error TS6133: 'assert' is declared but its value is never read.

20 import { assert } from 'chai';

apupier (changes requested):

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,197 @@
...
+
+const START_DEBUG_LABEL = uiTestConstants.startDebug;
+const REMOVE_INTEGRATION_LABEL = uiTestConstants.integrationRemove;
+
+describe.only('Test Debug on Camel K Integrations from Side Bar', function () {
+
+	before(async function () {
+		this.timeout(90000);
+		this.retries(3);
+		await VSBrowser.instance.waitForWorkbench;
```
apupier, line 50: Sonar is reporting: `Refactor this redundant 'await' on a non-promise.sonarlint(typescript:S4123)`

i think it shoudl be written `await VSBrowser.instance.waitForWorkbench()`

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,197 @@
...
+async function camelKToolingIsEnabled() {
+	const marketplace = await Marketplace.open();
+	const item = await marketplace.findExtension(`@installed ${pjson.displayName}`);
+	return extensionIsActivated(item);
+}
+
+async function moveToExplorerActivity() {
+	const ACTIVITY_LABEL = 'Explorer';
+
+	return await moveToActivity(ACTIVITY_LABEL);
```
apupier, line 127: redundant await (reported by Sonar)
```suggestion
	return moveToActivity(ACTIVITY_LABEL);
```

src/ui-test/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,197 @@
...
+	const section = await getCamelKIntegrationSection();
+	await VSBrowser.instance.driver.wait(async () => hasIntegration(integrationLabel));
+	const item = await section.findItem(integrationLabel) as ViewItem;
+	return item;
+}
+
+async function hasIntegration(integrationLabel: string) {
+	const section = await getCamelKIntegrationSection();
+	const isOnView = await sectionHasItem(section, integrationLabel);
+	return isOnView;
```
apupier, line 152: can return directly isOnView
```suggestion
	return await sectionHasItem(section, integrationLabel);
```

joshiraez:
Finallyyyyyyyyyiasdnkjankjesa

What is left to see if they pass also in the QE jenkins

src/ui-test/tests/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,198 @@
...
+				await createIntegration(INTEGRATION_FILE);
+				await startIntegration(INTEGRATION_LABEL);
+			})
+
+			it('Check Java Debug available', async function () {
+				this.timeout(20000);
+
+				const item = await getIntegration(INTEGRATION_LABEL);
+				const menu = await item.openContextMenu();
+				await promiseSucceed(VSBrowser.instance.driver.wait(() => menu.hasItem(START_DEBUG_LABEL), 5000));
```
apupier, line 74: the contextual menu must be closed when awaiting to have it with correct labels. here it iwll work only if the first time it is called it is already filled with the expected menu.

it will give something like:

```
await VSBrowser.instance.driver.wait(() => {
   const menu = await item.openContextMenu();
   const hasMenu = await menu.hasItem(SSTART_DEBUG_LABEL);
   menu.close();
   return hasMenu;
   });
```

src/ui-test/tests/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,198 @@
...
+		const WORKSPACE_FOLDERInput = await InputBox.create();
+		await WORKSPACE_FOLDERInput.selectQuickPick(0);
+		const nameInput = await InputBox.create();
+		await nameInput.setText(integrationFile);
+		await nameInput.confirm();
+
+		const editorView = new EditorView();
+		await VSBrowser.instance.driver.wait(async() => {
+			try {
+				return await editorView.openEditor(integrationFile + '.java') !== undefined;
```
apupier, line 169: Shouldn't the editor be opened when creating a new integration? Shouldn't we checked that the editor is opened instead of trying to open it?

joshiraez:
Oh no, tests are failing again after updating the configuration classes. I'm seeing if it's just some flaky test or something because my tests are passing and the others were passing well before, or I'll look into figuring what's happening

joshiraez:
Also there is a notification about the branch needs to be rebased but it's just today's merges. I'll get into that later

unsortedhashsets:
> Oh no, tests are failing again after updating the configuration classes. I'm seeing if it's just some flaky test or something because my tests are passing and the others were passing well before, or I'll look into figuring what's happening

Hi, @joshiraez 

Seems that test  breaks the extension (I do not know why for now), but result from screenshots:

image

unsortedhashsets:
> > Oh no, tests are failing again after updating the configuration classes. I'm seeing if it's just some flaky test or something because my tests are passing and the others were passing well before, or I'll look into figuring what's happening
> 
> Hi, @joshiraez
> 
> Seems that test breaks the extension (I do not know why for now), but result from screenshots:
> 
> image

With @joshiraez we have discussed the root of the problem (new folder open that caused reactivation of the extension)
So I have prepared small update to integrate test in the pipeline, for now seems to be stable 10/10 runs

joshiraez:
Thank you very much @unsortedhashsets for the help!!!! I don't think I'd have ever figured out that opening a directory would reset VSCode and much less that it would break the pipeline as the extension seemed to still be active and well when checking the VSCode testing window.

joshiraez:
Thank you very much @unsortedhashsets for the help!!!! I don't think I'd have ever figured out that opening a directory would reset VSCode and much less that it would break the pipeline as the extension seemed to still be active and well when checking the VSCode testing window.

djelinek (approved):

src/ui-test/tests/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,92 @@
...
+				this.timeout(uiTestConstants.TIMEOUT_60_SECONDS);
+				await createIntegration(INTEGRATION_FILE, 'Java', 'java');
+				await startIntegration(INTEGRATION_LABEL, 'Basic');
+			})
+
+			it('Check Java Debug available', async function () {
+				this.timeout(uiTestConstants.TIMEOUT_30_SECONDS);
+				const item = await getIntegration(INTEGRATION_LABEL);
+				const menu = await item.openContextMenu();
+				await promiseSucceed(VSBrowser.instance.driver.wait(() => menu.hasItem(START_DEBUG_LABEL), uiTestConstants.TIMEOUT_5_SECONDS));
```
apupier, line 50: Why not directly:

```
await VSBrowser.instance.driver.wait(() => menu.hasItem(START_DEBUG_LABEL), uiTestConstants.TIMEOUT_5_SECONDS)
```

src/ui-test/tests/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,92 @@
...
+				await createIntegration(INTEGRATION_FILE, 'Java', 'java');
+				await modifyCurrentFileToBeInvalid();
+				await startIntegration(INTEGRATION_LABEL, 'Basic');
+			});
+
+			it('Test Java Debugger Not Available On Invalid File', async function () {
+				this.timeout(uiTestConstants.TIMEOUT_30_SECONDS);
+				const item = await getIntegration(INTEGRATION_LABEL);
+				const menu = await item.openContextMenu();
+				await promiseFail(VSBrowser.instance.driver.wait(() => menu.hasItem(START_DEBUG_LABEL), uiTestConstants.TIMEOUT_5_SECONDS));
```
apupier, line 79: Why not directly

```
VSBrowser.instance.driver.wait(() => !menu.hasItem(START_DEBUG_LABEL), uiTestConstants.TIMEOUT_5_SECONDS)
```

src/ui-test/utils/waitConditions.ts
```diff
@@ -77,7 +78,22 @@ export async function viewHasItem(content: ViewContent, section: string, item: s
...
 }
 
-export async function updateFileText(oldText: string, newText: string): Promise<boolean> {
+export async function sectionHasItem(section: ViewSection, item: string, timePeriod = 2000): Promise<boolean> {
+    try {
+        const currentItem = await section.findItem(item);
+        if (currentItem !== undefined) {
+            return true;
+        } else {
+            await section.getDriver().sleep(timePeriod);
```
apupier, line 87: what is the purpose of sleeping before returning false?

apupier (changes requested):

src/ui-test/tests/camelk_java_debug_test.ts
joshiraez: First because that's not an assertion but counting on the code to work or fail whenever there is an exception.

Second because there are cases where promises could succeed or fail without having properly asserted the condition. We want to know if the wait command successfully completed and not just go forward.

There was more logic to it but I have to check the discussion because this was prompted by someone that requested it.

src/ui-test/tests/camelk_java_debug_test.ts
joshiraez: First because that's not an assertion but counting on the code to work or fail whenever there is an exception.

Second because there are cases where promises could succeed or fail without having properly asserted the condition. We want to know if the wait command successfully completed and not just go forward.

There was more logic to it but I have to check the discussion because this was prompted by someone that requested it.

src/ui-test/utils/waitConditions.ts
joshiraez: This code was refactored from somewhere else, so I'm not sure what was the logic behind it.

src/ui-test/tests/camelk_java_debug_test.ts
joshiraez: Specially this one is impossible to assert if we use the wait command without using this promiseFail function.

apupier: > First because that's not an assertion but counting on the code to work or fail whenever there is an exception.

the promiseFail is also not an assertion. it is based on the occurence of an Exception or not

> We want to know if the wait command successfully completed and not just go forward.

it is the case too with the

```
await VSBrowser.instance.driver.wait(() => !menu.hasItem(START_DEBUG_LABEL), uiTestConstants.TIMEOUT_5_SECONDS)
```

> Specially this one is impossible to assert if we use the wait command without using this promiseFail function.

i do not understand why. In Both cases are not using assertion for the failure.

If we want to have an assertion to provide more details on the reason of the failure than a timeout exception, it is possible to do something like:

```
try {
await VSBrowser.instance.driver.wait(() => !menu.hasItem(START_DEBUG_LABEL), uiTestConstants.TIMEOUT_5_SECONDS)
} catch (error) {
assert.fail('The start label should not be visible', e)
 }
```

src/ui-test/utils/waitConditions.ts
apupier: Can you find from where? In This PR, this code sounds to be new, i do not see any places from where it was deleted.

djelinek: from my understanding it will provide pause between next attempt of doing something.. so if we are waiting for something you can try every second or... try.. wait eg 3sec and try again

apupier: this code is not called with several attempts.
if we want this behavior i think it should be handled directly inside this method and using a conditional wait

unsortedhashsets: Hi all,

That function used by function from utils:
https://github.com/camel-tooling/vscode-camelk/blob/38e4065ebe0c82c8d88b06359443393975953422/src/ui-test/utils/utils.ts#L125-L128

which is used in:

https://github.com/camel-tooling/vscode-camelk/blob/38e4065ebe0c82c8d88b06359443393975953422/src/ui-test/utils/utils.ts#L119-L123

https://github.com/camel-tooling/vscode-camelk/blob/38e4065ebe0c82c8d88b06359443393975953422/src/ui-test/utils/utils.ts#L130-L137

It can be replaced by an already tested function I think:

https://github.com/camel-tooling/vscode-camelk/blob/38e4065ebe0c82c8d88b06359443393975953422/src/ui-test/utils/waitConditions.ts#L71-L79

which is used here:
https://github.com/camel-tooling/vscode-camelk/blob/38e4065ebe0c82c8d88b06359443393975953422/src/ui-test/tests/dev_mode_test.ts#L96-L100

and here

https://github.com/camel-tooling/vscode-camelk/blob/38e4065ebe0c82c8d88b06359443393975953422/src/ui-test/tests/basic_mode_test.ts#L82-L86

src/ui-test/tests/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,198 @@
...
+
+	async function moveToExplorerActivity() {
+		const ACTIVITY_LABEL = 'Explorer';
+
+		return moveToActivity(ACTIVITY_LABEL);
+	}
+
+	async function moveToActivity(viewIdentifier: string) {
+		const control = await new ActivityBar().getViewControl(viewIdentifier);
+		return VSBrowser.instance.driver.wait(async () => control?.openView());
```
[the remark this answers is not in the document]
apupier: the problem is that it will fail later on because the setup wasn't performed correctly but we won't know that. if it is failing at this place, we will know that the setup was not performed succesfully

src/ui-test/tests/camelk_java_debug_test.ts
apupier: There is interest in UI Tests which are more end to end test to be closer to expected behavior from users too.
it won't change the potential flakyness.
it will fail this test in case the command is no more working that's right.
it slightly complexifies the codebase to add this code.

Both ways have there advantages and drawbacks. I tend to privilege to be closer to real user scenario for UI tests but that's tfine to privilege decoupling.

src/ui-test/tests/camelk_java_debug_test.ts
joshiraez: Which is actually a similar version to the definition of promiseFail function

const assertPromise = async (prom: any, succed = true) => {
    succed = succed ? true : false;
    try {
        prom = Array.isArray(prom) ? prom : [prom];
        await Promise.all(prom);
        assert.strictEqual(true, succed);
    }
    catch (ex: any) { assert.notStrictEqual(true, succed, ex); }
}

But more general and better named.

I can add another param to the assertPromise function to add what the message should be in case it doesn't work. Would that work it out?

src/ui-test/utils/waitConditions.ts
joshiraez: >  if we want this behavior i think it should be handled directly inside this method and using a conditional wait

I strongly believe we should refrain from having explicit code doing explicit things inside of tests and leaving them to more general setup tests and compose our tests from those general utils.

> @unsortedhashsets  said

I'll try changing it to the already tested function you mentioned, thanks!

joshiraez: @unsortedhashsets I have been checking the functions, which underlay use

```
 export async function getCamelKIntegrationSection() {
    const section = await new SideBarView().getContent().getSection(uiTestConstants.extensionName) as CustomTreeSection;
    await section.expand();
    await VSBrowser.instance.driver.wait(async () => section.isExpanded());
    return section;
}
 ```
 
 As I had some failures because the section wasn't expanded so it couldn't find the items. From there, I went with the above function of "sectionHasItem" because it had the best interface for what we were using and sending the view content (sideBar) again seemed wasteful and misleading.
 
 I can work on refactoring the utils functions if we want but I have a feel that that might be outside of the scope of this PR.

File: src/ui-test/tests/camelk_java_debug_test.ts [hunk 27]
```diff
@@ -0,0 +1,98 @@
...
+			})
+
+			it('Check Java Debug available', async function () {
+				this.timeout(uiTestConstants.TIMEOUT_30_SECONDS);
+				const item = await getIntegration(INTEGRATION_LABEL);
+				const menu = await item.openContextMenu();
+				await assertPromiseSucceed(
+					VSBrowser.instance.driver.wait(() => menu.hasItem(START_DEBUG_LABEL), uiTestConstants.TIMEOUT_5_SECONDS),
+					'Java debug was not available on a run of a valid file'
+				);
```
apupier, line 53: this is working only if the menu item is available before opening the context menu. The context menu is not refreshed automatically in the background after displayed. So this is equivalent to:

```
const menu = await item.openContextMenu();
await menu.hasItem(START_DEBUG_LABEL);
```

to retry we need to open and close the contextual menu each time, something like:

```
await VSBrowser.instance.driver.wait(() => 
 {
      const menu = await item.openContextMenu();
      const hasMenu = await menu.hasItem(START_DEBUG_LABEL);
     menu.close();
     return hasMenu;
}
```

src/ui-test/tests/camelk_java_debug_test.ts
```diff
@@ -0,0 +1,98 @@
...
+			});
+
+			it('Test Java Debugger Not Available On Invalid File', async function () {
+				this.timeout(uiTestConstants.TIMEOUT_30_SECONDS);
+				const item = await getIntegration(INTEGRATION_LABEL);
+				const menu = await item.openContextMenu();
+				await assertPromiseFail(
+					VSBrowser.instance.driver.wait(() => menu.hasItem(START_DEBUG_LABEL), uiTestConstants.TIMEOUT_5_SECONDS),
+					'Java debugger was available for a run on an invalid file'
+				);
```
apupier, line 85: this is checking only directly after the contextual menu is opened. The context menu is not refreshed automatically in the background after displayed. So this is currently equivalent to something like:

```
const menu = await item.openContextMenu();
assertFalse(await menu.hasItem(START_DEBUG_LABEL));
```

I think that is needed is something like:

```
				await assertPromiseFail(
					VSBrowser.instance.driver.wait(() => 
{ 
const menu = await item.openContextMenu();
const hasMenu = menu.hasItem(START_DEBUG_LABEL);
await menu.close();
return hasMenu;
}, uiTestConstants.TIMEOUT_5_SECONDS),
					'Java debugger was available for a run on an invalid file'
				);
```

apupier (changes requested):

src/ui-test/tests/camelk_java_debug_test.ts
joshiraez: Oh, I see. Thanks!

src/ui-test/tests/camelk_java_debug_test.ts
joshiraez: Same as above then, I'll fix this. Thanks!

src/ui-test/tests/camelk_java_debug_test.ts
apupier: feedback not adressed

src/ui-test/tests/camelk_java_debug_test.ts
apupier: feedback not adressed

apupier (changes requested):

apupier:
closing as a bunch of changes occurred and there is now several conflicts an dno plan to work on it on short term

PR #1263, opened 2022-09-14
FUSETOOLS2-1799 - replace download package

apupier:
- download package is no more maintained and contains a security vulnerability through its chain of dependency
- using node-fetch 2.x despite 3.x is available. Unfortunately, it is not working as-is due to ESM module. 2.x is maintained for security fix and important bugs
- lost the functionality which was sharing progress of download. i have not found a maintained library capable of it and wasn't able to implement it with node-fetch and tar packages. I think this can be done in another iteration anyway. It is more important to have a secure extension. See https://issues.redhat.com/browse/FUSETOOLS2-1806 to potentially add it back


joshiraez (approved):

joshiraez:
_lost the functionality which was sharing progress of download. i have not found a maintained library capable of it and wasn't able to implement it with node-fetch and tar packages. I think this can be done in another iteration anyway. It is more important to have a secure extension._


Is there a ticket for this already? Would be cool to share a link in the first message in the PR for traceability

apupier:
> Is there a ticket for this already?

https://issues.redhat.com/browse/FUSETOOLS2-1806

> Would be cool to share a link in the first message in the PR for traceability

updated first message

PR #1280, opened 2022-09-28
FUSETOOLS2-1826 - use Camel K 1.10.0 by default

apupier:


src/kamel.ts
```diff
@@ -120,14 +128,17 @@ async function kamelInternal(command: string, devMode: boolean, namespace : stri
 	});
 }
 
-async function kamelInternalArgs(args: string[], devMode: boolean, namespace: string | undefined, foldername?: string): Promise<ChildProcess> {
+async function kamelInternalArgs(args: string[], devMode: boolean, namespace: string | undefined, operatorId: string | undefined, foldername?: string): Promise<ChildProcess> {
```
joshiraez, line 131: Maybe better to use `operatorId?`like with foldername

joshiraez (approved):
Small nitpick but seems fine overall

src/kamel.ts
apupier: I tried to keep the same level of requirement than for `namespace`. But maybe better to use the ? operator for all. Long time that I have not read this code so do not remember the reasoning of not using it. Can be interesting to have a closer look effectively.

PR #1480, opened 2023-05-09
FUSETOOLS2-2061 Call refresh two times

unsortedhashsets:
Hi,

I have found that after the first `camelk.classpath.refresh` command config file is updated with the right dependencies, however, the `checkExpectedCompletion` function ignores it.

If dependencies are refreshed two times updated config file (SAME) will be recognized and `checkExpectedCompletion` pass without errors.

@apupier Do you have any ideas where I can search for the root of that problem?

But anyway, we can open an extra issue and unblock tests with that double refresh :thinking:


apupier (approved):
> Do you have any ideas where I can search for the root of that problem?

No idea

> But anyway, we can open an extra issue and unblock tests with that double refresh

agree

PR #1552, opened 2023-07-05
Update release process in contributing guide

apupier:
- CircleCI is no more used
- There is no more an Eclipse Che plugin registry, it was used for Eclipse Theia, now the main IDE is (VS) Code Server


unsortedhashsets (approved):
Hi @apupier, @djelinek is on PTO till 14.7.2023.
If you are ok with my Review here is it :-)

PR #1594, opened 2023-08-28
FUSETOOLS2-1897 - remove usage of deprecated Kamel init

apupier:
- it consists in removing the Create Integration file command and suggest users to Create Camel route file provided by VS Code Language Support for Apache Camel. Consequently, added the other extension as a dependency.
- tests are heavily modified because we cannot mock anymore inputbox opened by another extension. So instead of creating files in unit tests, we already have them in the test workspace. The case with created files is covered by UI tests
- Upgrade CI to use Java 17 instead of 11 as Language Support has been added as a dependency and requires it.
- It is an important step to support Camel K 2.x


djelinek (approved):

djelinek:
FYI @unsortedhashsets

PR #1616, opened 2023-09-20
FUSETOOLS2-1888 - remove dependency on `kamel local`

apupier:
- `kamel local` has been removed from kamel cli 2.x
- Adding requirement on JBang CLI to do the same action

let's start with a lot less ambitious change because after investigation it sounds moving things to a better way using JDT-LS capabilities (similar to what VS Code JBang is doing) and in VS Code language Support will be even larger than I initially thought.

### DIFF
--- a/src/JavaDependenciesManager.ts
+++ b/src/JavaDependenciesManager.ts
@@ -17,26 +17,25 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import * as vscode from 'vscode';
-import * as kamelCli from './kamel';
 import * as utils from './CamelKJSONUtils';
+import { execSync } from 'child_process';
 
 const PREFERENCE_KEY_JAVA_REFERENCED_LIBRARIES = "java.project.referencedLibraries";
 export const CAMEL_VERSION = "3.20.1";
 
 export async function initializeJavaDependenciesManager(context: vscode.ExtensionContext): Promise<void> {
-	const destination = parentDestinationFolderForDependencies(context);
+	const destination = destinationFolderForDependencies(context);
 	await downloadCommonCamelKJavaDependencies(context);
-	await initializeJavaSettingManagement(path.join(destination, 'dependencies'));
+	await initializeJavaSettingManagement(destination);
 }
 
 async function downloadJavaDependencies(context: vscode.ExtensionContext, groupId: string, artifactId: string, version: string) {
-	const destination = parentDestinationFolderForDependencies(context);
+	const destination = destinationFolderForDependencies(context);
 	const pomTemplate = context.asAbsolutePath(path.join('resources', 'maven-project', 'pom-to-copy-java-dependencies.xml'));
-	const directDependenciesDestination = path.join(destination, 'dependencies');
-	fs.mkdirSync(directDependenciesDestination, { recursive: true });
+	fs.mkdirSync(destination, { recursive: true });
 
 	const mvn = require('maven').create({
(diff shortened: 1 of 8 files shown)


apupier:
Camel JBang 4.0.1 released but still requires https://github.com/apache/camel/pull/11550

src/JavaDependenciesManager.ts
```diff
@@ -17,26 +17,25 @@
...
 
 async function downloadJavaDependencies(context: vscode.ExtensionContext, groupId: string, artifactId: string, version: string) {
-	const destination = parentDestinationFolderForDependencies(context);
+	const destination = destinationFolderForDependencies(context);
 	const pomTemplate = context.asAbsolutePath(path.join('resources', 'maven-project', 'pom-to-copy-java-dependencies.xml'));
-	const directDependenciesDestination = path.join(destination, 'dependencies');
-	fs.mkdirSync(directDependenciesDestination, { recursive: true });
+	fs.mkdirSync(destination, { recursive: true });
 
 	const mvn = require('maven').create({
```
apupier, line 37: potential improvement: remove completely maven for initial setup and use jbang with an inlined route

apupier:
moving forward after Camel JBang 4.0.1 available but existing test is now hitting this bug https://issues.apache.org/jira/browse/CAMEL-19917

apupier:
> existing test is now hitting this bug https://issues.apache.org/jira/browse/CAMEL-19917

adapted test and I guess we will live wit this regression for this iteration.

Note: there is a bit of mix between dependencies in 3.x and 4.x. And to configure the version to use, it needs to be configured globally with Camel JBang options. When moving to VS Code language support, it will be better and settings will be reused. So not spendin gmore time on these use cases.

test Fixture with speci@l chars/MyRouteBuilderWithAdditionalDependencies.java
```diff
@@ -1,9 +1,10 @@
 // camel-k: dependency=mvn:org.apache.commons:commons-math3:3.6.1
 
 import org.apache.camel.builder.RouteBuilder;
+import org.apache.commons.math3.util.ArithmeticUtils;
 
 public class MyRouteBuilderWithAdditionalDependencies extends RouteBuilder{
 	public void configure() {
-        ArithmeticUt
+        ArithmeticUtils dummy;
```
apupier, line 8: need to use a simpler use case due to regression between kamel local and Camel Jbang dependency copy: https://issues.apache.org/jira/browse/CAMEL-19917

src/JavaDependenciesManager.ts
```diff
@@ -74,20 +73,13 @@ export async function downloadSpecificCamelKJavaDependencies(
...
-		const kamelLocal = kamelCli.create();
-		const parentDestination = parentDestinationFolderForDependencies(context);
-		await clearDestinationFolder(mainOutputChannel, parentDestination);
-		const command = `local build --integration-directory "${parentDestination}" "${uri.path}"`;
+		const destination = destinationFolderForDependencies(context);
+		await clearDestinationFolder(mainOutputChannel, destination);
+		const command = `jbang camel@apache/camel dependency copy --output-directory="${destination}" "${uri.path}"`;
 		try {
-			await kamelLocal.invoke(command);
+			execSync(command);
```
apupier, line 80: when moving to VS Code Language support, it might be nicer to use a VS Code task so that user have access to log in case of problem

djelinek (approved):

